PR #6, opened 2017-09-27
Initialized project structure.

arekmano:
- Docker Compose file is at the root of the project; use 'docker-compose
up to start all 3 services'
- Each module of the project has its own Dockerfile (ml_service,
nlp_service, web_app)
- Each module exposes its own service to the host:
  - ml_service is exposed at port 3001
  - nlp_service is exposed at port 3002
  - web_app is exposed at port 3000
- All 3 microservices have been initialized as python flask applications
(will most certainly change)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,15 @@
+version: '3'
+
+services:
+  ml_service:
+    build: ./src/ml_service
+    ports:
+      - 3001:3001
+  nlp_service:
+    build: ./src/nlp_service
+    ports:
+      - 3002:3002
+  web_app:
+    build: ./src/web_app
+    ports:
+      - 3000:3000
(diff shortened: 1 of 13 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,15 @@
...
+  ml_service:
+    build: ./src/ml_service
+    ports:
+      - 3001:3001
+  nlp_service:
+    build: ./src/nlp_service
+    ports:
+      - 3002:3002
+  web_app:
+    build: ./src/web_app
```
lancelafontaine, line 13: We will most likely have a separate `web_client` and `web_server` in the future, but we'll change that once the need arises.

lancelafontaine:
Branch name and commit message not conforming to CONTRIBUTING.md :( will  be enforced in the future by CI once that is set up.

lancelafontaine (approved):

Issue #2, opened 2017-09-27
Chat Interface


dodoels:
*Description*
- As a user, I want a chat interface to interact with the system

*Scope of Work*
- [x] Create UI [3 man hours]
- [x] Implement necessary functions for further integration [3 man hours]
- [x] Configure continuous integration to run tests [3 man hours]

*Acceptance Criteria*
- Users can access the web application through browsers and see the UI.

*Story Points*
- 3

*Priority*
- High

*Risk*
- Low

*Demo requirement*
1. Open web browser
2. Access the web application via localhost
3. Verify that the chat interface has loaded error-free

**Documentation, files, diagrams, etc. relating to this feature:**

See uploaded screenshots in comments.

***Changes***:

No changes to the initial engineering solution.

naregeff:
Interface does what it needs to do, i.e. provides a chat interface for a user.
Can't wait to see it boosted on NLP + with a sick UI. 
Closing issue.

lancelafontaine:
![chatbot-ui-1](https://user-images.githubusercontent.com/1548285/31050503-a66362ea-a619-11e7-929c-a5e1d1c2976b.png)
![chatbot-ui-2](https://user-images.githubusercontent.com/1548285/31050504-aaac9e7a-a619-11e7-8de3-100a2d6f3bbb.png)

naregeff:
Please check UI on safari. It seems unstable. Thank you.

arekmano:
Please open a bug when bugs arise. Old user stories should remain closed unless their functionality was not delivered.

Also screenshots of the issue would help in debugging errors.

Issue #4, opened 2017-09-27
Access system online


dodoels:
*Description*
- As a user, I want to access the system on a website using any browser on any device

*Scope of Work*
- Set up domain (2h)
- Set up CI and CD (8h)
- Host the web application (set up HTTP server/proxy) (2h)

*Acceptance Criteria*
- User enters the public URL with a domain. User is shown our system's web application.
- User enters the URL on a mobile device. User is shown the systems web application properly formatted for their device.

*Story Points*
- 8

*Priority*
- Medium

*Risk*
- Low

arekmano:
This story is a natural extension of #2. This would involve obtaining and setting up a domain name for our web application, (with the possible addition of HTTPS support?)

naregeff:
Is it relevant to reference browser support? IMO, it would be an acceptance criteria that the system behave similarly across IE, Chrome, Firefox, Safari, Opera, and other browsers.

arekmano:
Sure it is. Though some of the browsers in that list don't support certain functionality related to PWA. Safari + Edge do not yet support service workers.

lancelafontaine:
- [x] Website currently accessible at https://capstone.cyberjustice.ca
- [x] Continuous deployment is enabled. This means that any changes in our master branch are tested and immediately update our site upon success.
- [x] I tested the site on (Desktop) Chrome, Firefox, Opera and (Mobile) Chrome, Firefox, and the core chat functionality is present. However, given that much of the UI and application functionality will be changing in future iterations, it may make more sense to reevaluate cross-browser support once we're further along into the web application development.

naregeff:
Beautiful, works as advertised.

naregeff:
Beautiful, works as advertised.

PR #32, opened 2017-09-29
[#2] Chat Interface

dodoels:
- Basic chat interface mostly for POC (lacking interaction)
- Basic VueJS setup, there are more libraries need to be integrated


src/web_app/chatbot/index.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>justiceai</title>
```
ghost: proceZeus*

To be changed through the whole PR. JusticeAI is not the name agreed upon by our stakeholder.

src/web_app/chatbot/src/components/Hello.vue
```diff
@@ -0,0 +1,83 @@
...
+    <button v-on:click="openChatBot">Start</button>
+  </div>
+</template>
+
+<script>
+export default {
+  name: 'welcome',
+  data () {
+    return {
+      msg: 'Welcome to JusticeAI! To get started, please tell us what you’d like to do.',
```
mihaiqc: proceZeus*

arekmano:
Do we want to commit to Webpack? I'm not very familiar with it. I know its powerful but it requires a lot of configuration. I defer to F.E. experts for their opinion.

dodoels:
@arekmano  Webpack is generic when you install VueJS. I do not think that using other build tools is a good idea.

dodoels:
Here is other build tool work with VueJS
https://github.com/sohelamin/vue-example

lancelafontaine:
@arekmano this webpack config was mostly generated by `vue-cli` which manages everything you would need. Unless we need to change something drastically, we shouldn't have to ever change these webpack config files. I agree that Webpack is definitely unwieldy and config-heavy, but unfortunately it's the future of everything related to front-end development. On the bright side, everything reasonable that we would want to do is already existing as an npm script.

lancelafontaine:
Since this is all static content,it should probably be in a separate directory with something like `nginx` like `web_client_proxy` or something. At the moment there's a Dockerfile for a web backend in this directory also. Let's sort out directory structure / dockerization of front-end in a future issue (like #4 for example).

src/web_app/chatbot/.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
```
lancelafontaine, line 1: Will need to make sure this aligns with our linting strategy

src/web_app/chatbot/package.json
```diff
@@ -0,0 +1,86 @@
+{
+  "name": "justiceai",
```
lancelafontaine: proceZeus

src/web_app/chatbot/.editorconfig
lancelafontaine: Actually, just realized this PR also contains our linting. Never mind 👍

src/web_app/chatbot/package.json
```diff
@@ -0,0 +1,86 @@
...
+  "version": "1.0.0",
+  "description": "Cyberjustice Laboratory Legal Chatbot",
+  "author": "CYberjustice Laboratory",
+  "private": true,
```
lancelafontaine: It's public so can remove this line

lancelafontaine (approved):
All configuration looks good, and running the application, tests and linting works as expected 👍

ghost (approved):

src/web_app/chatbot/.editorconfig
dodoels: I want to use airbnb linting in next iteration so will be more changes coming

PR #33, opened 2017-09-30
[#2] Dockerized the web client.

arekmano:
- Created a basic NodeJs docker container.
- Moved files from chatbot to root of webapp directory.


lancelafontaine:
Since our front-end is purely static, I was thinking we could eventually have a `web_client_and_http_proxy` docker image (containing with ☝️ and `nginx`), and a `web_server` image (containing our backend service). That's something we can do later (once we have a need for for a backend).

lancelafontaine (approved):

ghost (approved):

PR #38, opened 2017-10-05
[#4] Moves web_app to web_client, creates cjl script for docker managment

lancelafontaine:
[#4]

- [x] Renames the `web_app` directory to `web_client` to be more explicit
- [x] Creates `cjl`, a bash script to manage everything related to docker for all services
- [x] Configures continuous integration to adapt to above changes


cjl
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env bash
```
arekmano, line 1: lawyer :'(

arekmano: Also, I'm worried that a script like this will be unnecessary maintenance

.dockerignore
```diff
@@ -0,0 +1,2 @@
+./**/node_modules/
```
arekmano: disagree. each micro-service (docker image) should have its own .dockerignore. Shouldn't be root's problem

src/web_client/Dockerfile
```diff
@@ -4,10 +4,11 @@ WORKDIR /usr/src/app
 
 RUN apt-get install git -y
 
-COPY . .
-
+COPY src/web_client/package.json .
```
arekmano: this will prevent us from being able to build the web_client from its directory

arekmano (changes requested):

cjl
lancelafontaine: That's a possibility. If we find it to be too cumbersome in the future, I guess we can scrap it.

Now that it works though, adding new functionality should be very trivial. Should also be a huge time saver for most of us.

.dockerignore
lancelafontaine: Fixed

src/web_client/Dockerfile
lancelafontaine: Fixed

arekmano:
looks good but I absolutely thing that this will be a maintenance issue further down the line.

arekmano (approved):

Issue #20, opened 2017-09-28
Claim information gathering


dodoels:
*Description*
As a user, I would like the system to ask me questions to get information about my claim

*Scope of Work*
- [x] Create rough draft of categories and questions to ask the client (12 hours)
- [x] Construct decision tree for our bot (4 hours)
- [x] Research and confirmation with lawyer about specific questions and categories (5-20 hours) (Back and forth interaction)
- [x] Confirm questions bot need to ask the users (5 hours)
- [x] Confirm categories that the user could use the bot for (4 hours)
- [x] Confirm possible questions/responses per categories (25 hours)
- [x] Confirm decision tree for our bot (2 hours)

*Story Points*
- 20

*Priority*
- High

*Risk*
- High

*Demo requirement*
1. The user accesses the website from their machine
2. The bot starts the conversation by asking for the user's name
3. The user inputs their name
4. The bot asks the user whether they're a landlord or a tenant
5. The user inputs whether they're a landlord or a tenant
6. The bot asks the user to describe their dilemma
7. The user describes their dilemma using natural language 
8. The bot correctly identifies the category of dispute the dilemma belongs to by associating the dilemma to various articles of jurisprudence, asking for clarification if the bot cannot process the user's input
9. The bot begins asking questions relevant to the category of the dispute in order to build a claim

*Acceptance Criteria*
- ~~The bot is able to identify the type of aid the user requires~~
- ~~The bot is able to ask questions in reference to my particular subject with my particular issue~~
- The bot is able to gather the user's name, person type (landlord/tenant), and ask about their dilemma
- The bot is able to ask questions relevant to building a claim for the user's dilemma

**Documentation, files, diagrams, etc. relating to this feature:**
https://www.rdl.gouv.qc.ca/
http://legisquebec.gouv.qc.ca/fr/ShowDoc/cs/CCQ-1991
http://legisquebec.gouv.qc.ca/en/showdoc/cs/CCQ-1991
https://www.rdl.gouv.qc.ca/fr/etre-locataire/droits-et-obligations-du-locataire
https://www.rdl.gouv.qc.ca/fr/questions-frequentes
https://www.rdl.gouv.qc.ca/fr/etre-locataire/paiement-du-loyer
http://legisquebec.gouv.qc.ca/fr/ShowDoc/cr/R-8.1,%20r.%205 (in front of regie du logement)
[...]

PR #39, opened 2017-10-07
[#20] Adds backend service 

Vynny:
[#20]

- Adds flask backend_service as microservice
- Adds backend_service to docker-compose.yml with port 3003


src/backend_service/requirements.txt
```diff
@@ -0,0 +1,7 @@
+click==6.7
```
lancelafontaine, line 1: Don't think `click` will end up being required.

src/backend_service/setup.py
```diff
@@ -0,0 +1,10 @@
+from setuptools import setup
+
+setup(
```
lancelafontaine, line 3: Since we aren't publishing anything on PyPI, don't think this is required either. We can remove later once we're sure.

lancelafontaine (approved):
Good template to start with 👍

src/backend_service/setup.py
Vynny: It's there because I followed this guide http://flask.pocoo.org/docs/0.12/patterns/packages/

Issue #8, opened 2017-09-28
Chat with System (NLP)


dodoels:
As a user, I want to express an issue I'm facing using natural language

*Scope of work*
- [ ] Determine whether user input is relevant (12 hours)
- [ ] Clarify micro-service API endpoints (2 hours)
- [ ] In ambiguous situations, confirm input with user (1 hours)
- [ ] Determine the answer to a yes or no response to a question by the user (6 hours)
- [ ] Determine the classification of an answer into a given category (6 hours)
- [ ] Determine what the primary fact is within a response to a question by the user (15 hours)

*Acceptance Criteria*
- System asks a yes/no question. User gives a response describing a yes/no sentiment. System is able to classify the response among specific response types.
- System asks a question requiring a long answer. User gives a textual response. System is able to parse the response and classify the response among specific response types.

*Story Points*
- 20

*Priority*
- High

*Risk*
- High

lancelafontaine:
*Scope of work*
- [ ] Prompt user to enter name (? hours)
- [ ] Determine whether user input is relevant (? hours)
- [ ] In ambiguous situations, confirm input with user (? hours)
- [ ] Determine inconsistencies between responses (? hours)

lancelafontaine:
![20170929_193506](https://user-images.githubusercontent.com/1548285/31039682-a1bd3d32-a54d-11e7-8c67-7feeb89c00d9.jpg)

Samuel-Campbell:
-----------------------------------------
Standford Library
-----------------------------------------
PROS:
- Supports French and English
- Better runtime than NLTK

CONS:
- GPL license conflicts with our MIT license

-----------------------------------------
SpacY
-----------------------------------------
PROS:
- Better performance overall than NLTK. (Word tagging)

CONS:
- Only supports english

-----------------------------------------
Textblob
-----------------------------------------
Built over NLTK and pattern libraries

PROS:
- Better runtime than NLTK
- Good for very simple NLP operations
- Implements NLTK and pattern efficiently together

CONS:
- Functionality is more limited and less options for customization

-----------------------------------------
Gensim
-----------------------------------------
PROS: 
- Good for finding document similarities (not very useful to us)
- word2vec

CONS: 
- only proficient at finding document similarities and offers limited other functionaliteis
- GNU license

-----------------------------------------
Pattern
-----------------------------------------
PROS:
- Better runtime than NLTK
- Efficient at speech tagging and word tagging

CONS: 
[...]

PR #41, opened 2017-10-08
[#8] api endpoints

arekmano:
- Created classifiers for tenant/landlord and problem categories
- Added new endpoints
- Updated docker build process for this container


src/nlp_service/models/question.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def __init__(self, questionID, conversationID, answerString):
+        self.questionID = questionID
+        self.conversationID = conversationID
+        self.answerString = answerString
+        return
+
+
+class QuestionOutput(object):
+    """docstring for Question"""
```
lancelafontaine: Don't seem very useful for now

lancelafontaine (approved):
Looks good to me, but you should update your branch with changes from master.

arekmano:
updated from master

arekmano:
I'd like @Vynny, @mihaiqc or @Samuel-Campbell to take a look at the changes as this will impact their work.

src/nlp_service/services/introduction_parser.py
```diff
@@ -0,0 +1,18 @@
...
+    def classify(introduction):
+        personClass = 'unknown'
+        if introduction.personClassString == "test":
+            personClass = 'landlord'
+        else:
+            personClass = 'tenant'
+
+        return IntroductionOutput(
+            introduction.conversationID,
+            'Jackie Chan',
```
Vynny, line 16: This should be the person's name

Vynny (changes requested):

Samuel-Campbell:
Only comment I have is that I notice you don't use encapsulation. Other than that looks fine

src/nlp_service/services/introduction_parser.py
arekmano: This functionality isn't implemented yet. It returns a static string until we figure it out.

arekmano:
I didn't use encapsulation because it is easier to debug this way.

Samuel-Campbell (approved):

Vynny (approved):

PR #52, opened 2017-10-10
[#4] Adds volumes to development environment for autoreload, etc

lancelafontaine:
[#4]

- [x] Adds volumes to development environment for autoreload, etc.

I can confirm that the volumes actually work. The `web_client` service is able to develop and autoreload the latest changes. However, autoreload does not seem to be working for the Flask services. I added the `DEBUG=1` environment variable like the documentation says, but still no autoreload. That is something that we will have to revisit later.

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -4,16 +4,24 @@ services:
   ml_service:
     ports:
       - 3001:3001
+    volumes:
+      - ./src/ml_service:/usr/src/app
   nlp_service:
     ports:
       - 3002:3002
+    volumes:
+      - ./src/nlp_service:/usr/src/app
   backend_service:
     ports:
       - 3003:3003
+    volumes:
+      - ./src/backend_service:/usr/src/app
   web_client:
     command: npm start
     depends_on:
       - backend_service
     ports:
       - 3039:3039
+    volumes:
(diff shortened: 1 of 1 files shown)


dodoels (approved):

arekmano (changes requested):
copy command

docker-compose.dev.yml
```diff
@@ -4,16 +4,24 @@ services:
   ml_service:
     ports:
       - 3001:3001
+    volumes:
+      - ./src/ml_service:/usr/src/app
```
arekmano, line 8: we should remove the COPY command from the Dockerfiles in this case.

lancelafontaine: I don't want to make multiple different Dockerfiles for different environments though. Purpose was to keep everything identical except for the `docker-compose.dev.yml` file. @arekmano

arekmano: we're technically now copying and then overriding contents of folder with volumes. ohwell

arekmano (approved):

docker-compose.dev.yml
lancelafontaine: Yeah, I agree it's not the best, although I think it's an acceptable tradeoff for the reduced complexity.

Issue #15, opened 2017-09-28
Claim evidence gathering


dodoels:
*Description*
- As a user, I want the system to prompt me for the evidence required to support my claim

*Scope of Work*
- [x] A low-fid designs for system
- [x] A mockup for evidence-gathering 
- [x] Improve UI based of product owner mockup choice
- [x] Implement backend server with db
- [x] Validate what pieces of evidence are required by law-domain expert (too vague, split into #80, #84 for now)
- [x] Determine all possible valid piece of evidence to be supplied by the user (too vague, split into #25 #26)
- [x] Obtain image/text file/decision from the user and store in database

*Acceptance Criteria*
- [x] Chatbot prompts user to provide relevant evidence for case (generic question)
- [x] Chatbot is able to prompt for all possible relevant evidence in a variety of scenarios [CANCELLED] (too vague, split into #80 , #84 )
- [x] User is able to provide evidence in textual and image-based formats
- [x] Chatbot is able to retrieve evidence in textual and image-based formats

*Story Points*
- 13

*Priority*
- High

*Risk*
- High

arekmano:
- [ ] 3 UI designs for system (3h ideal time)
- [ ] 2 UI Mockups for evidence-gathering (2h ideal time)
- [ ] Improve UI based of product owner mockup choice (3h)
- [ ] Implement backend server with db (12h)
- [ ] Validate what pieces of evidence are required by law-domain expert (2h)
- [ ] Determine all possible valid piece of evidence to be supplied by the user (5h)
- [ ] Obtain image/text file/decision from the user and store in database (3h)
- [ ] Have a basic yes/no decision tree of what evidence to ask for (? h)

arekmano:
As part of this user story, we should finalize the domain of law that we will be diving into

arekmano:
UI Designs for both the system and the feature should be completed by this thursday, so that we can obtain feedback about it on friday.

arekmano:
Acceptance criteria "Chatbot is able to prompt for all possible relevant evidence in a variety of scenarios" is overly broad. We have to limit this user story's scope.

arekmano:
This story may also involve the creation and persistance of data in a database, that is capable of storing various types of data. Also "Have a basic yes/no decision tree of what evidence to ask for (? h)" is being worked on as a side-effect of #20. 

For "Validate what pieces of evidence are required by law-domain expert", we can limit this to images of a lease and receipts

naregeff:
[...]

PR #53, opened 2017-10-11
[#15] PostgreSQL database and CI environment

lancelafontaine:
[#15]

- [x] Updates docs in all READMEs for how to run tests and lints for a particular service
- [x] Adds a CI `docker-compose` to simplify running tests and lints
- [x] Sets up a Docker network for ability to resolve hostnames across Docker containers
- [x] Sets up a Docker volume for database persistence when Docker containers are teared down
- [x] Creates a PostgreSQL database as a service provided by a container
- [x] Develops a proof-of-concept for running queries against the database from the backend service while both are running in separate containers using `docker-compose`.

### DIFF
--- a/docker-compose.ci.yml
+++ b/docker-compose.ci.yml
@@ -1,0 +1,14 @@
+version: '3'
+
+services:
+  ml_service:
+    command: bash -c "pytest && flake8"
+  nlp_service:
+    command: bash -c "pytest && flake8"
+  backend_service:
+    command: bash -c "pytest && flake8"
+  web_client:
+    command: bash -c "npm run test && npm run lint"
+  postgresql_db:
+    environment:
+      POSTGRES_PASSWORD: DEV_PASS_NOT_SECRET
--- a/src/postgresql_db/README.md
+++ b/src/postgresql_db/README.md
@@ -1,0 +1,18 @@
+# PostgreSQL Database
+
+## Connect via command line
+
+Bring up all services:
+
+```
+./cjl up -d
+```
+
+Connect via `psql`:
+```
+./cjl run --rm postgresql_db "psql -h postgresql_db -U postgres"
+```
+
+The above command will prompt you to enter the database password.
+
+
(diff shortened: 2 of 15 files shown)


dodoels (approved):

docker-compose.ci.yml
```diff
@@ -0,0 +1,14 @@
...
+    command: bash -c "pytest && flake8"
+  nlp_service:
+    command: bash -c "pytest && flake8"
+  backend_service:
+    command: bash -c "pytest && flake8"
+  web_client:
+    command: bash -c "npm run test && npm run lint"
+  postgresql_db:
+    environment:
+      POSTGRES_PASSWORD: DEV_PASS_NOT_SECRET
```
arekmano, line 14: What about our prod pass?

src/postgresql_db/README.md
```diff
@@ -0,0 +1,18 @@
+# PostgreSQL Database
```
arekmano, line 1: I just want to make sure that PROD data is not ephemeral; we most certainly need user input to further train our classifiers.

arekmano (changes requested):
I would like my prod-related question answered before merge.

docker-compose.ci.yml
lancelafontaine: Was thinking of GPG encrypted password decryptes with a key that is only the on prod machine.

src/postgresql_db/README.md
lancelafontaine: Dev data is not ephemeral either. It's using a volume to store on your local. For prod data, we will have to include some kind of db backup strategy.

arekmano (approved):
Approved but I would like some form of plan for the prod database

lancelafontaine:
Will do. I'll start Prod configuration soon. I'll propose something, but I'll also speak with everyone on the team to determine the needs for the db for all services. It will give me a better idea of what is required.

PR #64, opened 2017-10-13
[#20] claim information gathering

Vynny:
- Adds backend service API
- Adds integration with nlp_service
- Bot will determine whether person is landlord or tenant
- Bot will determine category of question
- Bot will ask questions to resolve facts for category until none are left, then print fact dump


src/backend_service/models/models.py
```diff
@@ -0,0 +1,101 @@
+import datetime
```
arekmano, line 1: i prefer the SQLAlchemy models split into different files. in a later iteration

CONTRIBUTING.md
```diff
@@ -66,7 +66,7 @@ An example might be:
 [#12] Added user login feature
 
     - Created user model
-    - Added Oauth library to our dependencies for authentication
+    - Added Oauth library to our dependencies for authyaoentication
```
arekmano, line 69: we don't actually have auth though

src/backend_service/controllers/conversationController.py
```diff
@@ -0,0 +1,141 @@
+from flask import jsonify, abort, make_response
```
arekmano, line 1: please keep naming scheme consistent. Either camelCase or snake_case, not both

arekmano:
my review is about cosmetic things, or rather errors that would be cool if they were fixed.

arekmano (approved):

src/backend_service/controllers/conversationController.py
Vynny: I used camel for modules, pascal for classes and snake for methods

src/backend_service/controllers/conversationController.py
```diff
@@ -0,0 +1,141 @@
...
+
+from models.factService import FactService
+from models.models import *
```
TaimoorRana, line 4: just curious: why can you just import models.models instead? since you importing everything either way.

Vynny: models is the package (folder models/)
models.models is is the file (models/models.py)
there are several db models in models.py represented by classes, so I import them all with *

PR #77, opened 2017-10-16
[#15] Adds service dependencies to base docker configuration in order prevent failure when spinning up services.

lancelafontaine:
[#15]

- [x] Adds service dependencies to base docker configuration in order  to prevent failure when spinning up services.
- [x] Sets log level in gunicorn to DEBUG to show Flask debug messages

This prevents failures on `./cjl up` where the backend_service loads and attempts to connect before the database service has spun up.

### DIFF
--- a/docker-compose.base.yml
+++ b/docker-compose.base.yml
@@ -4,21 +4,30 @@ services:
   ml_service:
     build: ./src/ml_service
     image: procezeus/ml_service
+    depends_on:
+      - postgresql_db
     networks:
       - procezeus_net
   nlp_service:
     build: ./src/nlp_service
     image: procezeus/nlp_service
+    depends_on:
+      - postgresql_db
     networks:
       - procezeus_net
   backend_service:
     build: ./src/backend_service
     image: procezeus/backend_service
+    depends_on:
+      - postgresql_db
+      - nlp_service
     networks:
       - procezeus_net
   web_client:
     build: ./src/web_client
     image: procezeus/web_client
+    depends_on:
+      - backend_service
     networks:
       - procezeus_net
   postgresql_db:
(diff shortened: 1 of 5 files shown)


lancelafontaine:
According to Docker documentation, this fix is not actually sufficient to resolve the problem.

https://docs.docker.com/compose/startup-order/

When using `depends_on`, `docker-compose` will not wait until a container is “ready” (whatever that means for your particular application) - only until it’s running. 

I'll need to write some small "health check" scripts to make sure that the database is actually running and ready to accept connections before spinning up the `backend_service`.

arekmano:
a hack that I had used back in Docker 12 was to have a bash script execute and curl the required services until they responded, I would then run the service that depended on them. i.e.

A depends on B, C

for i = 1 to 20:
    curl B
    if $? = 0 break
    if i = 20 exit(1) // timeout
    sleep(1 second)

for i = 1 to 20:
    curl C
    if $? = 0 break
    if i = 20 exit(1) // timeout
    sleep(1 second)

execute process A

arekmano:
Is this PR still being worked on?

dodoels:
Is this PR still valid ? @lancelafontaine

lancelafontaine:
Yes. Changes incoming.

TaimoorRana (approved):

lancelafontaine:
I'm rather confident that I've inadvertently solved this problem by using Gunicorn.

- Default Gunicorn settings is to kill and restart any worker thatt has been idle for more than 30 seconds. Theoretically,  if the backend service started before the DB service, it would break, be killed 30 seconds later, and then be restarted (once DB service had already been started)
- Gunicorn _seems_ to fork a process on bootup, but not actually run the flask application until required (ie. an initial request comes in). Note that I have not confirmed this, this is just my suscpicion.

I'm very confident that the issue is solved by ☝️ . In the case that it is not, I'll create another PR to resolve the issue at that point.

lancelafontaine:
@arekmano

arekmano (commented):
LGTM

arekmano (approved):

PR #78, opened 2017-10-16
[#15] Allows Docker containers to write to host directories

lancelafontaine:
[#15]

- [x] Reintroduce shared volumes between Docker container and host
- [x] Allows Docker containers to write to host directories

During development, the containers are reading from and writing to the host's file system. 

The implication of reading from the host filesystem is that fancy development features like auto-watching files and hot module reloading now works from within the Docker container.

The implication of writing to the host filesystem is that the Docker container is now able to write data files or generate dependencies that are currently required for running the `web_client` and `nlp_service` services.

### DIFF
--- a/cjl
+++ b/cjl
@@ -28,6 +28,8 @@ if [ -z "$COMPOSE_FILE" ]; then
 fi
 
 # Passes script arguments to docker-compose
+sudo find ./src -type d -exec chmod 777 {} \;
+sudo find ./src -type f -exec chmod 666 {} \;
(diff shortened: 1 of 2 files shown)


lancelafontaine:
Before being merged, I'll need the sign off from either @choitwao or @Vynny that this solution actually works under Windows. You can test this in the following manner:

1. Make sure that `node_modules` does not exist in `src/web_client`
2. `./cjl up`
3. Make sure that web_service is running as expected on `http://localhost:3039` and that there are no errors relating to requiring `chalk` as a dependency in the console output
4. `node_modules` now exists in the `src/web_client` directory

dodoels (approved):

arekmano (approved):

PR #89, opened 2017-10-19
[#15] Provides functions for automatically linting all Python+Javascript files

lancelafontaine:
[#15]

- [x] Creates functionality in `cjl` script to automatically fix any linting errors throughout the codebase.

Note that CI will still fail if there are linting errors. All you have to do is before you push, do `./cjl lint-fix`. You can even add that as a git pre-commit hook if you like to ensure that your build never fails because of linting errors.


dodoels (approved):

src/nlp_service/services/gram_classifier.py
```diff
@@ -26,8 +26,13 @@ class GramClassifier(object):
     of the given input"""
 
     def __init__(self, baseName, inputFiles, forceTrain):
-        if os.path.exists(os.getcwd() + '/data/' + baseName + '/' + baseName + '.pickle') \
-                and not forceTrain:
+        if os.path.exists(
```
arekmano: this is why I'm not too happy about obligatory linting...

src/backend_service/models/factService.py
```diff
@@ -29,7 +29,8 @@ class FactService:
     @staticmethod
     def get_question(claim_category, facts_resolved):
         all_category_facts = list(FactService.fact_dict[claim_category])
-        facts_unresolved = [x for x in all_category_facts if x not in facts_resolved]
+        facts_unresolved = [
+            x for x in all_category_facts if x not in facts_resolved]
```
arekmano: what? how is this better :joy:

Vynny (approved):
Approved since the line length rules were removed and silly things don't happen anymore

PR #90, opened 2017-10-20
[#15] Python service autoreload with Gunicorn

lancelafontaine:
[#15]
- [x] Sets up Gunicorn, a production-ready gateway for Flask applications. Uses 3 workers due to 1 core being available on our production VMs.
- [x] Sets up autoreload for all Flask-based services so that development is much quicker.


arekmano:
We should make sure that our back-end does not keep state. I'm unsure if it does so currently (postgres caching, ID generation, conversation state) but having 3 workers can lead to odd bugs if we do maintain it

Vynny (approved):
Will certainly speed up development. 
Backend currently only stores state in DB so it shouldn't be a problem. In the future, we could always use redis if we need caching or state persistence state outside of DB.

PR #93, opened 2017-10-25
[#15] Adds backend file upload API

Vynny:
- [x] Adds API endpoint for uploading files to conversation
- [x] Adds API endpoint for getting conversation's file metadata
- [x] Adds API endpoint for retrieving file by id
- [x] Adds file requests in messages (optional)


src/backend_service/README.md
```diff
@@ -129,11 +157,106 @@ Gets the message history for a conversation
...
+
+**Content-Disposition** : attachment; filename=file_name.extension
+
+**Body** : Binary file content
+
+## Error Response
+
+**Code** : `400 Bad Request`
+
+**Code** : `404 Not Found`
```
lancelafontaine, line 267: Loving these API docs

src/backend_service/services/fileService.py
```diff
@@ -0,0 +1,41 @@
...
+
+from werkzeug.utils import secure_filename
+
+from app import app
+
+UPLOAD_FOLDER = 'uploads'
+ALLOWED_EXTENSIONS = {'pdf', 'png', 'jpg', 'jpeg'}
+
+app.config['UPLOAD_FOLDER'] = UPLOAD_FOLDER
+app.config['MAX_CONTENT_LENGTH'] = 10 * 1024 * 1024
```
lancelafontaine: Do we ever handle the case where a file is too big?

src/backend_service/services/fileService.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def upload_file(file, file_path, filename):
```
lancelafontaine, line 15: Some of these could definitely be tested.

lancelafontaine (commented):
This is awesome stuff! 👍 

Only comment is that once we deal with user accounts, we'll have to make sure we retroactively fix some of the `conversationController` stuff to check for authentication. IIRC Flask has some nice decorators for this kind of stuff.

lancelafontaine (approved):

src/backend_service/app_test.py
```diff
@@ -1,11 +1,20 @@
 from models.factService import FactService
```
Samuel-Campbell: Shouldn't you use unittest for this?

src/backend_service/controllers/conversationController.py
```diff
@@ -29,37 +34,111 @@ def init_conversation(name):
...
+        return abort(make_response(jsonify(message="File does not exist"), 404))
+    else:
+        abort(make_response(jsonify(message="Conversation does not exist"), 404))
+
+
+##################
+# Private Methods
+##################
+
 def _generate_response(conversation, message):
```
Samuel-Campbell: Your methods are Protected and Not private.
use: __ for private and _ for protected

src/backend_service/services/fileService.py
Vynny: Flask will just return an error in that case.

Samuel-Campbell (changes requested):
I suggest using more Object oriented as opposed to functional programming. At your discretion

src/backend_service/services/fileService.py
Vynny: It's not super straightforward to create flasks 'file' object to test with but I'll look into it today

src/backend_service/services/fileService.py
```diff
@@ -0,0 +1,41 @@
...
+    # Create the directory
+    if not os.path.exists(file_path):
```
Samuel-Campbell, line 17: I think you should transform this file into OOP

src/backend_service/controllers/conversationController.py
```diff
@@ -92,12 +172,22 @@ def _determine_person_type(conversation, message):
...
+        elif person_type == 'landlord':
+            response = StaticStrings.chooseFrom(StaticStrings.problem_inquiry_landlord).format(name=conversation.name)
+
+        return {'response_text': response, 'file_request': file_request}
     else:
-        return StaticStrings.chooseFrom(StaticStrings.clarify)
+        return {'response_text': StaticStrings.chooseFrom(StaticStrings.clarify)}
 
 
 def _determine_claim_category(conversation, message):
```
Samuel-Campbell: protected, not private

src/backend_service/controllers/conversationController.py
```diff
@@ -1,11 +1,16 @@
...
 from models.models import *
 from models.staticStrings import *
-from services import nlpService
+from services import nlpService, fileService
 
 
+########################
+# Conversation Handling
+########################
+
```
Samuel-Campbell, line 13: Make this OOP

Samuel-Campbell (changes requested):

src/backend_service/services/fileService.py
lancelafontaine: I personally disagree. Free functions in a module are perfectly fine for this use case.

Samuel-Campbell: Yeah this file is the only one I wasn't sure about.

dodoels (approved):

src/backend_service/controllers/conversationController.py
Vynny: What would the benefit of making this OOP be?
conversationController is essentially just a module that prevents app.py being clogged up with too much logic

src/backend_service/controllers/conversationController.py
Vynny: My bad, I'll change these and other instances of the same error.

src/backend_service/controllers/conversationController.py
Samuel-Campbell: Ok I didn't realize it was an extension of app.py

src/backend_service/README.md
```diff
@@ -53,13 +65,26 @@ Provide the conversation_id and a message. Message should be empty string for fi
...
 }
 ```
 
+Response containing a request for a file.
+
+```json
+{
+    "conversation_id": 1,
+    "file_request": {
+        "document_type": "LEASE"
```
arekmano, line 83: can we document the "document_type" enum somewhere?

src/backend_service/controllers/conversationController.py
```diff
@@ -29,42 +34,116 @@ def init_conversation(name):
...
+                    fileService.get_file_extension(file), fileService.get_accepted_formats_string())), 400))
+
+    else:
+        abort(make_response(jsonify(message="Conversation does not exist"), 404))
+
+
+def get_file(conversation_id, file_id):
+    conversation = Conversation.query.get(conversation_id)
+
+    if conversation:
```
arekmano: Can we abstract these "if conversation ,else 404" clauses

src/backend_service/README.md
arekmano: :+1:

arekmano (changes requested):

src/backend_service/controllers/conversationController.py
Vynny: Any suggestion on how?
Eventually we might not just be querying by primary key, so it seemed cleaner to just do the query at the beginning of the method and return a 404 if it isn't found. 
In my opinion it might be overkill to abstract away a simple if else statement.

arekmano: I was thinking it may be worth not repeating ourselves for each endpoint. With something along the lines of:
def getConversationOrAbort(id):
conv = Conv.get(id)
if conv:
return conv
else:
abort()

arekmano: but this refactor is not a showstopper IMO. simply something I would add

arekmano (approved):

Samuel-Campbell (approved):

PR #94, opened 2017-10-25
[#5] eula interface

dodoels:
- Create functional UI interface with 1) chat interaction 2) chat history 3) sidebar with file list
- Basic VI

image
image
image
image


lancelafontaine:
The `Proce` and `eus` in the logo are not centered (vertically or horizontally) and my OCD is kicking in.

src/web_client/src/components/Chat.vue
```diff
@@ -4,34 +4,84 @@
...
-    <b-row id="chat-title">
-      <b-col sm="2"><h2>ProceZeus</h2></b-col>
-    </b-row>
-    <div id="chat-widow" v-chat-scroll>
-      <div class="chat-message-container" v-for="(message, index) in chatLog">
+    <!-- Chat History -->
+    <transition name="fade">
+      <div id="chat-history" v-if="openChatHistory" v-chat-scroll>
+        <ul>
+          <li v-for="(conv, index) in chatLog">
```
lancelafontaine: not using the `index` here

src/web_client/src/components/Chat.vue
```diff
@@ -74,38 +125,31 @@ export default {
         message: this.currentUserInput
       }).then(
         response => {
-          this.showMessage(this.currentUserInput, 'USER');
-          this.showMessage(response.body.message, 'BOT');
+          this.currentZeusInput = null;
+          setTimeout(() => {
```
lancelafontaine, line 129: Why `setTimeout` here?

src/web_client/src/components/Chat.vue
```diff
@@ -41,14 +91,15 @@ export default {
     return {
       chatLog: new Array,
       currentUserInput: null,
-      username: 'Hammer of Justice!', // TODO: prompt user's name
-      connectionError: false
+      currentZeusInput: null,
+      username: 'Bruce Lee', // TODO: prompt user's name
```
lancelafontaine: We should track this TODO somewhere.

src/web_client/src/components/Chat.vue
```diff
@@ -74,38 +125,31 @@ export default {
...
         },
         response => {
           this.connectionError = true;
         }
       );
     },
-    getChatHistory (zeusId) {
+    getChatHistory () {
+      let zeusId = this.$localStorage.get('zeusId');
       this.$http.get('http://localhost:3003/conversation/' + zeusId).then(
```
lancelafontaine, line 141: This hardcoded `localhost:3003` will break in Docker. It should be `http://backend_service/conversation`.

URL should also probably be in a config somewhere to prevent hard-coding it everywhere.

src/web_client/src/components/Sidebar.vue
```diff
@@ -0,0 +1,57 @@
...
+
+<template>
+    <div id="sidebar-component">
+        <div id="sidebar-logo">
+            <img alt="" src="../assets/logo.png">
+            <p>Alpha</p>
+        </div>
+        <div id="sidebar-account">
+            <img alt="" src="../assets/user_avatar_1.png">
+            <h2>BRUCE CAI</h2>
```
lancelafontaine: Hardcoding here too. We should either thinking about passing props into components (depending on the hierarchy), or defining a `global_values.js` which has values that are used across every component.

lancelafontaine: Or introducing state management library like Redux... Depends on how much complexity we want to take on right away.

lancelafontaine (approved):
I left a few comments that I think are important to address, but they are not showstoppers. We should definitely track them to make sure they get done though.

src/web_client/src/components/Chat.vue
dodoels: For amimation

src/web_client/src/components/Sidebar.vue
dodoels: This actually should be the `username`, and `username` should be defined in landing page or something. Also this area will be implemented when we create account feature. I do not think vuex (vue redux) is useful here but increasing the complexity of the code... The front-end logic is almost all here.

src/web_client/src/components/Chat.vue
dodoels: For the logo, I will adjust it in next PR.

Issue #68, opened 2017-10-14
Be able to make the distinction between original case and rectified case


ghost:
*Description*
- As a user, I want to see which cases were rectified.

*Scope of Work*
- [x] Be able to process text files
- [x] Be able to detect whether a case had multiple hearings
- [x] Be able to make the distinction between the original case and a rectified case with their different "No dossier"
- [x] Be able to parse a user-defined amount of randomly selected cases
- [x] Save extracted data in JSON format

*Acceptance Criteria*
- The user is able to see which cases had multiple hearings

*Demo requirement*
1. The user chooses a set amount of cases to randomly verify if actually rectified.
2. The system provides the set amount of cases and clearly identifies which ones have been rectified **with** the case's title (textfile title) and its rectification status, "No dossier" (both rectified and original) as well as "No demande"

*Story Points*
- 20

*Priority*
- High

*Risk*
-  High

**Documentation, files, diagrams, etc. relating to this feature:**
Url, books, etc.

***Changes***:
- [Optional]

naregeff:
The demo requirement will show that the functionality is working. The implementation is different: the system should, as a general rule, not rely on data from cases that have been appealed, since a higher court could be saying the literal opposite of the lower court ruling. 

The basic user doesn't need to know these distinctions; just what's valid caselaw at the end of the day.

Unverified, priority high.

arekmano:
Our hypothesis was that, since a higher court's decision overrules a lower one, that the information found in the appeal is a more accurate representation of the case. As a prerequisite to using the appropriate data (i.e. the fact as it is described within the appeal, which contradicts the fact as it is described within the original case), we must be able to distinguish between the various "phases" of a claim.

This ability to distinguish is described in the current user story.

naregeff:
In light of @arekmano comment, verified.

naregeff:
Great, this is done. Create new user story for appeals, etc. once data is available.

List of all situations where a decision could be overthrown: https://www.rdl.gouv.qc.ca/fr/decision/contestation-d-une-decision

PR #95, opened 2017-10-25
[#68] detect rectified cases

TaimoorRana:
**Description**
As a user, I want to see which cases were rectified, appealed, etc.

**Task completed:** 

1.  Process text files in a given directory
2.  Detects whether a case had multiple hearings
3.  Differentiates between the original case and a rectified case with their different "No dossier"
4.  Parse a user-defined amount of randomly selected cases
5.  Save extracted data in JSON format

**Sample JSON output:**
`{"noDossier": ["123456 12 12345678 D"], "noDemande": ["1234567"], "isRectified": true, "totalHearings": 2}`


src/ml_service/feature_extraction/dummy_cases/sample_case1.txt
```diff
@@ -0,0 +1 @@
...





No dossier:
"NODOSSIER"
No demande:
"NODEMANDE"
Page 2
```
TaimoorRana: This is fake data. All names, dates, and amounts are fake with most text replaced with lorem lipsum.

arekmano (approved):
LGTM

src/ml_service/feature_extraction/case.py
```diff
@@ -0,0 +1,73 @@
...
+
+import re
+import codecs
+import json
+import os
+
+
+class Case:
+    def __init__(self, file_path):
+        self.data = {"noDossier": 0, "noDemande": 0, "isRectified": False, "totalHearings": 0}
```
Vynny: Do the keys have to be French? Why not recordNumber, caseNumber..etc?

Vynny (approved):
Looks good to me, was only confused with the French keys names

src/ml_service/feature_extraction/case.py
TaimoorRana: They can be in English for sure, I have no preference. I’ll change it.

Samuel-Campbell: yes english plz

PR #97, opened 2017-10-26
[#15] fix download file issue

dodoels:
- fix the file download issue: now can download file directly by calling the end point. (instead of creating file in local)
- Border color change when push message


arekmano:
What is this change about? May we have a short description? the code and commit message aren't self-explanatory

Vynny (approved):
Looks good, can eventually change it to use BLOB + file saver to take care of edge cases where the file might've been deleted.

PR #98, opened 2017-10-26
[#5] Adds addditional prompt for disclaimer acceptance with related functionality

lancelafontaine:
[#5]

- Adds a disclaimer prompt to the conversation between bot and user.
- Populates response with related information (`additional_info`, `possible_answers`)

Next steps:
- [ ] Front-end: disable text input, display answer buttons
- [ ] Front-end: parse message for disclaimer as HTML based on `additional_info.should_parse_as_html`
- [ ] Front-end: Create legal page which fetches entire legal text from backend and displays it
- [ ] Backend: Create a `GET /legal` endpoint which stores and versions legal documents.


src/backend_service/controllers/conversationController.py
```diff
@@ -35,9 +35,13 @@ def receive_message(conversation_id, message):
     conversation = __get_conversation(conversation_id)
 
     file_request = None
+    possible_answers = None
+    additional_info = None
     # First message in the conversation
     if len(conversation.messages) == 0:
-        response_text = StaticStrings.chooseFrom(StaticStrings.welcome).format(name=conversation.name)
+        response_text = StaticStrings.chooseFrom(StaticStrings.disclaimer).format(name=conversation.name)
+        possible_answers = ["Yes"]
```
Vynny: Shouldn't there be a "No" answer?

lancelafontaine: A couple of us discussed this a bit, and felt like if the user absolutely disagreed with the disclaimer, they could just not use the service, or close the browser tab.

arekmano:
Good initial implementation

src/backend_service/services/staticStrings.py
```diff
@@ -2,6 +2,11 @@
 
 
 class StaticStrings:
+    # Asking for Acceptance to legal conditions
+    disclaimer = [
+        "Hello {name}! Before we start, I want make it clear that I am not a replacement for a lawyer and any information I provide you with is not meant to be construed as legal advice. Always check in with your legal professional. You can read more about our terms of use <a href='/legal'>here</a>. Do you accept these conditions?"
```
TaimoorRana: minor typo in line 7: I want to*

src/backend_service/controllers/conversationController.py
```diff
@@ -35,9 +35,13 @@ def receive_message(conversation_id, message):
...
+        additional_info = {"should_parse_as_html": True}
```
TaimoorRana: just curious, what are we parsing as HTML?

lancelafontaine: We should parse the disclaimer text as HTML, because it will contain an HTML link to a legal page.

src/backend_service/controllers/conversationController.py
```diff
@@ -133,6 +143,9 @@ def __get_conversation(conversation_id):
 
 
 def __generate_response(conversation, message):
+    if len(conversation.messages) <= 2:
```
TaimoorRana: Why is 2 being used?  what does 2 represent?

Vynny: Wondering the same, this could be a bit more descriptive tbh

src/backend_service/controllers/conversationController.py
Vynny: Instead of additional_info, it'd be better if it returned a 'response_html' rather than 'response_text', that way its a simple if else on the front end

lancelafontaine: Yeah, that makes sense. I'll change that in the next commit.

src/backend_service/controllers/conversationController.py
lancelafontaine: Will make more explicit in next commit.

src/backend_service/controllers/conversationController.py
```diff
@@ -61,10 +72,16 @@ def receive_message(conversation_id, message):
     db.session.commit()
 
     # Build response dict
-    response_dict = {'conversation_id': conversation.id, 'message': response_text}
+    response_dict = {'conversation_id': conversation.id}
 
+    if response_text is not None:
+        response_dict['message'] = response_text
+    if response_text is not None:
```
Vynny: This should be response_html

Vynny (changes requested):

src/backend_service/controllers/conversationController.py
lancelafontaine: Fixed in  0bf4305

Vynny (approved):

dodoels (approved):

PR #104, opened 2017-10-27
[66] Sets up a minimal production docker-compose configuration prod env

lancelafontaine:
[#6]

- [x] Sets up a minimal production docker-compose configuration


docker-compose.prod.yml
```diff
@@ -0,0 +1,19 @@
...
+      - 3002:3002
+  backend_service:
+    ports:
+      - 3003:3003
+  web_client:
+    command: npm start
+    ports:
+      - 3039:3039
+  postgresql_db:
+    environment:
```
arekmano, line 18: - expose : "PORT NUMBER"

https://docs.docker.com/compose/compose-file/#expose

arekmano (changes requested):

arekmano (approved):

docker-compose.prod.yml
lancelafontaine: docker-compose.dev.yml has no port number and has been working for weeks. Clearly it's not required.

arekmano:
postgres port exposure is enabled in the Dockerfile provided by maintainers. It is not required to add the EXPOSE command for linking.

Issue #66, opened 2017-10-13
Persist conversation text


lancelafontaine:
*Description*
As a user, I would like my conservations to be persisted.

*Scope of Work*
- [ ] Expose port in production
- [ ] Have production environment
- [ ] Figure out how to conveniently access prod data

*Story Points*
- 5

*Priority*
- High

*Risk*
- Low

*Demo requirement*
1. User chats with the bot
2. User can see the history of this conversation (in a database)

*Acceptance Criteria*
- User enters piece of conversation which is found in database
- Conversation can be retrieved for future analysis 

**Documentation, files, diagrams, etc. relating to this feature:**
Url, books, etc.

***Changes***:
- [Optional]

lancelafontaine:
Blocked: needs a privacy policy / EULA. Issue: #5

ghost:
Blocked: Somewhat confused between distinction between this issue and #23

naregeff:
Unverified, don't follow the use of this user story. Please explain.

Edit: verified.
Edit2: priority high

arekmano:
In order to perform a prediction about the user's claim, the system requires information gained from the user. This information needs to be persisted within our system at the very least temporarily (even if it is removed after a set time) for us to make a prediction.

This will also allow for the improvement to our chatbot's classifiers

arekmano:
priority? From an architectural point of view, this is fairly critical for development to continue on the system.

naregeff:
Does what it says it should do. Good work!

PR #108, opened 2017-10-27
[#66] Always read DB password from environment variable, for both backend and DB services

lancelafontaine:
[#66]

- [x] Sets default passwords for the database as environment variables for the dev and CI docker-compose environments
- [x] Sets prod docker-compose environment to read environment variable from the host
- [x] Uses the production node.js configuration for web client in production environment


dodoels (approved):
looks legit

Issue #105, opened 2017-10-27
Fact clustering


arekmano:
*Description*
As a user, I would like the system to determine categories of facts based on precedent data.

Major work will be conducted on #105. The same approach will then be used for #106.

*Scope of Work*
- [x] Preprocess data of decisions. 
- [x] Extract all facts from all obtained precedents.
- [x] Cluster facts based on sentence similarity.
- [x] Create a text file for each cluster of facts where the similar facts are saved, for manual review.

*Story Points*
- 8

*Priority*
- High

*Risk*
- High

*Acceptance Criteria*
- Given a fact, the system should return the cluster label for it.
- Looking at the label's text file, the written facts should be similar to the input fact.

arekmano:
For the moment, my test script in unsupervised-play is using Meanshift with the python library sk-learn
http://scikit-learn.org/stable/modules/clustering.html

Here is what is done so far:
1. Parse facts from each Precedent (needs major improvement, it truncates some facts half-way)
2. Vectorize each fact using word2vec on each word, and then averaging the for all words in a sentence  (https://www.youtube.com/watch?v=ERibwqs9p38)
3. Use the sentence vectors to cluster each fact within clusters using Meanshift (http://scikit-learn.org/stable/modules/generated/sklearn.cluster.MeanShift.html#sklearn.cluster.MeanShift)
4. Output all the sentences into files.

arekmano:
What can be improved:

- We are vectorizing and averaging the sentences. A better approach, according to (https://stackoverflow.com/questions/29760935/how-to-get-vector-for-a-sentence-from-the-word2vec-of-tokens-in-sentence) is to take the TF-IDF into account as well.
- We are using non-lemma version to vectorize words (http://embeddings.org/frWac_non_lem_no_postag_no_phrase_200_cbow_cut100.bin) a better approach can be to use a lemmatized version of the word2vec model and lemmatize our words.
- Regex to extract facts must not truncate half a fact
- At the moment facts AND outcomes are regexed out, we should limit to JUST facts so that we can cluster them ONLY.
- Experiment with different clustering algorithms to see which is the better one for our use case. http://scikit-learn.org/stable/modules/clustering.html
- Save the clustering algorithm to a pickle file so that we don't have to retrain the model every time we start the system (https://docs.python.org/2/library/pickle.html)
- Add labels to precedent JSON files

Samuel-Campbell:
Preprocessing
------------------------
[...]

PR #118, opened 2017-11-01
[#105] Added fact clustering to the ML Service

arekmano:
[#105] 

- Clustering uses DBSCAN algorithm
- Facts are vectorized using word2vec, trained on
  French non-lemmatized internet web crawler data.
- Test supervised learning script added
- Added init script to download required files.


src/ml_service/feature_extraction/fact_clustering/fact_clusterer.py
```diff
@@ -0,0 +1,27 @@
...
+    labels = ms.labels_
+    n_clusters = len(np.unique(labels))
+    print("Number of estimated clusters : %d" % n_clusters)
+    writeFactsToFile(factDict, labels)
+
+
+def writeFactsToFile(factDict, labels):
+    outputDirPath = os.path.join(os.path.dirname(
+        os.path.realpath(__file__)), '../../outputs/')
+    for i, sent in enumerate(factDict.keys()):
```
Samuel-Campbell, line 22: maybe it's just in the view on git but i don't see indent under your for loop for the body. IT would suck to train 8 hours and catch an error here

arekmano: updated

Samuel-Campbell (approved):

PR #122, opened 2017-11-01
[#105] Added model downloading script to init.py

arekmano:
- Required model is now downloaded to the root of ml service folder
  when init.py is run

[#105]


Samuel-Campbell (approved):

src/ml_service/init.py
```diff
@@ -1,4 +1,28 @@
...
+from threading import Thread
 
 nltk.download('punkt')
 nltk.download('stopwords')
+
+
+def monitorDownload(filename, filesize):
+    time.sleep(2)
+    current_size = 0
+    limit = filesize * 0.99
```
TaimoorRana, line 15: why 0.99?

Samuel-Campbell: I think it's just because he's updating the user with the progress of the download. after 99% it's complete. In any case can we push this so i can use it? por favor

TaimoorRana: oh I see. There should be some comments IMO

arekmano:
Assumption was that after 99% we don't need to let the user know about progress since it is near complete. Also saves 3 second sleep time.

arekmano:
Are you ok with this being merged? I'll add comments in a later PR

Issue #112, opened 2017-10-27
Choose user type before conversation


lancelafontaine:
*Description*
As a user I would like to indicate my user type on a landing page

*Scope of Work*
- [x] Create landing page mock-up  
- [x] Create landing page with user type selection
- [x] Pass user type to dashboard 
- [x] Remove file download on front-end
- [x] Initial conversation session based on the user type
- [x] Remove question of asking user type
- [x] Remove file download, get file list api logic

*Story Points*
- 5

*Priority*
- Medium

*Risk*
- Low

*Demo requirement*
1. The user should see the landing page 
2. The user should be able to select from either tenant or landlord
3. The system should start a conversation related to the user type

*Acceptance Criteria*
1. The user should see the landing page 
2. The user should be able to select from either tenant or landlord
3. The system should start a conversation related to the user type

***Changes***:

dodoels:
![image](https://user-images.githubusercontent.com/12192439/32157395-8eca7706-bd19-11e7-9cdf-1da8f30cb734.png)
This is my suggested logic flow for this story. The UI will be designed based on this flow. I also considered the extensibility of the design for the account story. The new flow will be based on the above one.
![image](https://user-images.githubusercontent.com/12192439/32157429-d212cd6a-bd19-11e7-812e-e29b149da2af.png)

dodoels:
Landing page design (med-fid)
![image](https://user-images.githubusercontent.com/12192439/32201540-1ceec606-bdae-11e7-9ca9-e6bf1ca09919.png)

dodoels:
We should also remove file download, get file list apis in this story

choitwao:
#

Vynny:
Get file metadata endpoint will remain in order to provide an upload trace. File downloading capabilities were removed.

naregeff:
It works!

PR #124, opened 2017-11-02
[#112] Implement Element UI and fix Node compatibility

dodoels:
- Bootstrap is not compatible with Node 8 and up, therefore refactor the UI using Element UI
- Therefore fix the bootstrap issue when building 
- Fix travis config when revert commit is not recognized and thus build fails
- Fix docker node version


Samuel-Campbell (approved):

.travis.yml
```diff
@@ -7,7 +7,7 @@ services:
 env:
   - CI_ENV=`bash <(curl -s https://codecov.io/env)`
 install:
-  - BRANCH=$(echo $TRAVIS_BRANCH | egrep  "^([0-9]+/|master)"); if [ ! $BRANCH ]; then exit 1; fi;
+  - BRANCH=$(echo $TRAVIS_BRANCH | egrep  "([0-9]+/|master)"); if [ ! $BRANCH ]; then exit 1; fi;
```
TaimoorRana, line 10: Why is this being changed?

dodoels: When you have revert the commit name will be "revert [#XXX]" which you can't change but Travis is going to fail the build immediately

PR #125, opened 2017-11-02
[#112] Changes person type resolution flow

Vynny:
[#112]
- [x] Changes conversation creation endpoint to require both 'name' and 'person_type' attributes
- [x] Asking whether user is tenant/landlord no longer part of conversation. Accepting EULA will lead immediately to problem inquiry.

### DIFF
--- a/src/backend_service/controllers/conversationController.py
+++ b/src/backend_service/controllers/conversationController.py
@@ -18,8 +18,11 @@ def get_conversation(conversation_id):
     return ConversationSchema().jsonify(conversation)
 
 
-def init_conversation(name):
-    conversation = Conversation(name=name)
+def init_conversation(name, person_type):
+    if person_type.upper() not in PersonType.__members__:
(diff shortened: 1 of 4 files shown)


src/backend_service/controllers/conversationController.py
```diff
@@ -18,8 +18,11 @@ def get_conversation(conversation_id):
     return ConversationSchema().jsonify(conversation)
 
 
-def init_conversation(name):
-    conversation = Conversation(name=name)
+def init_conversation(name, person_type):
+    if person_type.upper() not in PersonType.__members__:
```
dodoels, line 22: So `person_type` can be lower or upper case?

dodoels (approved):
looking good

src/backend_service/controllers/conversationController.py
Vynny: yes, either is fine

PR #126, opened 2017-11-03
[#112] Removes file download API

Vynny:
[#112]
- [x] Files can no longer be downloaded
- [x] File metadata can still be retrieved to have an upload trace


Vynny:
Commit 708dc0e "Removes file upload API' was a typo. Amended to 9ac996e "Removes file download API".

dodoels (approved):

PR #127, opened 2017-11-03
[#112] finalize anonymous landing page interaction

dodoels:
- fix bunch of css using element UI. better responsive now (not complete, but way better)
- now file list is still there, but read only (no download is allowed)
- initial chat session based on user type, no tenant/landlord question any more


src/web_client/src/components/Landing.vue
```diff
@@ -45,6 +45,7 @@ export default {
     },
     methods: {
         chooseType(usertype) {
+            this.$localStorage.set('username', usertype) //TODO: prompt username
```
Vynny, line 48: I don't think the username and the person's name are going to be the same.
The person's name might be "Bob Smith", but when they make their account their username would be bob-smith or bobsmith..etc.

dodoels: `username` is referring `user's name` imo. For anonymous users I want to prompt them to enter their actual names; for registration, I want to have their actual name, email, and password. That being said no traditional `username` will be used.

Vynny: When we do implement accounts, I don't think our usernames should have spaces. Perhaps if we only accepted emails as account names this would be avoided.

Vynny (approved):

src/web_client/src/components/Landing.vue
dodoels: Yes, should be just `email` and `password` alone with `fullname`

PR #128, opened 2017-11-03
105/retrain ner

Samuel-Campbell:
- Retrained the ner with more data
- Added stop words


src/ml_service/preprocessing/French/GlobalVariable.py
```diff
@@ -13,7 +13,23 @@ def get_path():
 
 def get_stop_words():
     try:
```
arekmano: why do you try/except here?

src/ml_service/preprocessing/main.py
```diff
@@ -12,9 +12,9 @@
     parser = Parser()
     j = 0
     for i in os.listdir(Global.Precedence_Directory):
-        if 'AZ-51210' in i and j < 5:
+        if 'AZ-5120' in i and j < 10:
```
arekmano, line 15: what is this magic string changing?

Samuel-Campbell: it's the same string you use to find texts in the decision bank. I'm just experimenting with different texts at random

src/ml_service/preprocessing/French/GlobalVariable.py
Samuel-Campbell: removed

src/ml_service/preprocessing/French/DecisionParse.py
```diff
@@ -121,5 +121,5 @@ def __populate_model(self, line):
     # ** This method can be enhanced for better classification
     #    This is just a proof of concept for now
     def __split_sub_sentence(self, sentence):
-        sentence = sentence.replace(',', ".")
+        #sentence = sentence.replace(',', ".")
```
arekmano, line 124: remove unused line?

arekmano (approved):

PR #129, opened 2017-11-03
[#112] Ensures that cjl gets rebuilt without a cache in production during CD

lancelafontaine:
[#112]
- [x] Ensures that cjl gets rebuilt without a cache in production during CD

This will likely be refactored soon, as we may head towards using an Ansible playbook instead of this not-so-nice oneliner.


dodoels:
@lancelafontaine the issue number ref and in title is wrong, it is #112

lancelafontaine:
Thanks, fixed now

dodoels (approved):

Vynny (approved):
It'll take a bit longer, but definitely worth it to avoid errors.

arekmano (approved):

PR #131, opened 2017-11-04
[#112] Attempts to install ansible from Ubuntu repositories instead of from a ppa

lancelafontaine:
- [x] Attempts to install ansible from Ubuntu repositories instead of from a ppa


lancelafontaine:
#112

TaimoorRana (approved):

PR #146, opened 2017-11-09
[#105] new tf-idf which is trained on the 45k precedents

TaimoorRana:
new tf-idf which is trained on the 45k precedents


dodoels (approved):

Samuel-Campbell (approved):

Samuel-Campbell:
Please comment out the code i told you in this PR

Samuel-Campbell (changes requested):
Comment out the manifold process in this pr / delete the line
remove try/except for hdbscan import in main.py

TaimoorRana:
@Samuel-Campbell we shouldn't remove the try and catch because it's a safeguard when running on a machine which doesn't have HDBSCAN prereqs.  it's currently running on VMs, so we don't to worry about it.

Samuel-Campbell (approved):

Issue #75, opened 2017-10-14
Better Classification for Tenant/Landlord Problem


arekmano:
*Description*
As a user, I would like the problem classification to have a higher precision and recall, so that I can properly classify my response

*Scope of Work*
- [x] Re-implement nlp service system using RASA
- [x] Research improvements to classifying text in NLP
- [x] Augment current data on classification of Tenant/Landlord's problem for "Deposit" category
- [x] Augment current data on classification of Tenant/Landlord's problem for "Rent Change" category
- [x] Augment current data on classification of Tenant/Landlord's problem for "Lease Termination" category
- [x] Augment current data on classification of Tenant/Landlord's problem for "Non Payment" category
- [x] Implement classification improvement

*Story Points*
- 8

*Priority*
- Medium

*Risk*
- Medium

*Acceptance Criteria*
- System classifies the Tenant/landlord's problem with an F1 score of 85% or higher

naregeff:
Verified. Realistically, I believe this one will be at a medium priority, as we need the system to improve its classification. Keep in mind that depending on what accuracy threshold we're able achieve with the initial story, this may change to become a high (or low) priority item.

arekmano:
Updated

arekmano:
Just checked out RASA NLU library, which is a wrapper around a suite of NLP tools. This may come in handy for NLP as it will have a better performance than a single technique

Vynny:
New persistence model structure 
![20171103_182230_1024](https://user-images.githubusercontent.com/1418403/32455314-e2d45822-c2ef-11e7-9d6e-06c5b6703c43.jpg)

Vynny:
System communication flowchart, to be followed when implementing new NLP system.

![20171103_182253](https://user-images.githubusercontent.com/1418403/32455358-090c0ae4-c2f0-11e7-8197-7d2f3aec3aaf.jpg)

Vynny:
Proposed ER diagram. I was thinking that the Fact table could represent the existence of a particular fact in the system, meaning if we had 20 total facts that could be resolved, there'd be 20 columns.
FactEntity would have a foreign key to the Fact, and would signify a resolved fact with a determined value. 

![erdiagram](https://user-images.githubusercontent.com/1418403/32587712-f2eccc5a-c4d7-11e7-8658-a12ebbd717e9.PNG)

lancelafontaine:
Looks good to me, but every fact would be stored as a row (not as a column like you mentioned), right?

[...]

PR #148, opened 2017-11-09
[#75] Replaces NLP system with Rasa

Vynny:
[#75]

*This is a work in progress*

- [x] Imports rasa NLU into nlp service
- [x] Adds training data for problem classification and fact extraction
- [x] Adds fact extraction & problem classification via rasa NLU engine
- [x] Defines new HTTP interfaces between backend -> nlp
- [x] Changes db models to new format (now with FactEntities)
- [x] Changes application data flow to agreed upon format (backend <-> nlp *<->ml*)


src/nlp_service/app.py
```diff
@@ -1,59 +1,47 @@
...
-def find_answer():
-    return None
-    # query_json = request.get_json()
-    # query = query_json['query']
-    # statement = query_json['statement']
-    # answer = Answer()
-    # answer_model = AnswerModel(answer.query(query, statement))
-    # return jsonify(answer_model.__dict__)
+##This is the RASA Trainer
+training_data = load_data('rasa/data/Problem_identifier.json')  # Where to fish the data it is being trained
```
arekmano: This training should be done in init.py or equivalent. Else, initializing will take forever

src/nlp_service/rasa/app_test.py
```diff
@@ -0,0 +1,20 @@
...
+from rasa_nlu.converters import load_data
+from rasa_nlu.model import Trainer
+from rasa_nlu.model import Metadata, Interpreter
+from rasa_nlu.components import ComponentBuilder
+
+##This is the RASA Trainer
+start_time = time.localtime(time.time())
+training_data = load_data('data/is_habitable.json')
+trainer = Trainer(RasaNLUConfig("config/config_spacy.json"))
+trainer.train(training_data)
```
arekmano: Has potential to break travis when training data becomes sufficiently large

src/nlp_service/rasa/data/data.json
```diff
@@ -0,0 +1,320 @@
+{
+  "rasa_nlu_data": {
```
arekmano: sample data does not belong in our project

arekmano (dismissed):

src/nlp_service/app.py
ghost: We will be changing a lot of the structure shortly.

src/nlp_service/rasa/data/data.json
ghost: This will also be removed.

src/backend_service/app.py
```diff
@@ -1,14 +1,17 @@
...
+import util
+
+util.load_src_dir_to_sys_path()
+from postgresql_db import database
 
 app = Flask(__name__)
 
 # DB Setup
 db = database.connect(app, 'postgres', os.environ['POSTGRES_PASSWORD'], 'postgres')
+#db = database.connect(app, 'postgres', 'postgres', 'postgres', host='127.0.0.1')
```
arekmano: no commented code

src/postgresql_db/models.py
```diff
@@ -0,0 +1,192 @@
...
+    possible_answers = db.Column(db.Text)
+    enforce_possible_answer = db.Column(db.Boolean)
+
+    # One to one
+    relevant_fact = db.relationship('Fact', uselist=False, backref='message')
+    file_request = db.relationship('FileRequest', uselist=False, backref='message')
+
+    def request_file(self, document_type):
+        file_request = FileRequest(document_type=document_type)
+        self.file_request = file_request
```
Samuel-Campbell: not a big deal for this one but consider that 'file_request' is declared static in this scope and using the keyword self. you are creating a separate instance of file_request.
I don't know if it's intentional

src/postgresql_db/models.py
```diff
@@ -0,0 +1,192 @@
...
+    id = db.Column(db.Integer, primary_key=True)
+
+    # Foreign Keys
+    message_id = db.Column(db.Integer, db.ForeignKey('message.id'))
+
+    # Attributes
+    document_type = db.Column(db.Enum(DocumentType), nullable=False)
+
+    def __init__(self, document_type):
+        self.document_type = document_type
```
Samuel-Campbell: same as previous comment

src/nlp_service/rasa/rasa_classifier.py
```diff
@@ -0,0 +1,37 @@
+import os
+
+from rasa_nlu.config import RasaNLUConfig
+from rasa_nlu.converters import load_data
+from rasa_nlu.model import Trainer, Interpreter
+
+
+class RasaClassifier(object):
```
Samuel-Campbell: why are you inheriting object?

src/nlp_service/rasa/app_test.py
```diff
@@ -0,0 +1,36 @@
...
+
+    interpreters[fact_key] = Interpreter.load(model_directory, RasaNLUConfig)
+
+print(interpreters['is_student'].parse(u"I am a student."))
+print(interpreters['is_habitable'].parse(u"It smells pretty bad but overall it's a livable place."))
+print(interpreters['has_lease_expired'].parse(u"Yes my lease expired."))
+print(interpreters['lease_type'].parse(u"My lease ends in June."))
+
+# {'intent': {'confidence': 0.65315696536207235, 'name': 'ask_lease_termination'},
+#  'entities': [{'value': 'faggot', 'entity': 'person', 'start': 3, 'end': 9, 'extractor': 'ner_crf'},
```
Samuel-Campbell: NICE VALUE! 👍

src/nlp_service/rasa/rasa_classifier.py
Vynny: My bad, forgot it was implicit.

src/postgresql_db/models.py
Vynny: Yeah it's creating a new file request. I guess it could be simplified to self.file_request = FileRequest...

src/nlp_service/controllers/nlpController.py
```diff
@@ -0,0 +1,116 @@
+from flask import jsonify, abort, make_response
+
+# from postgresql_db.models import *
```
arekmano: remove?

src/nlp_service/controllers/nlpController.py
```diff
@@ -0,0 +1,116 @@
...
+    question = None
+
+    fact_entity_value = __extract_entity(current_fact, message)
+    if fact_entity_value is not None:
+        # Pass fact with extracted entity to ML service
+        new_fact = mlService.submit_resolved_fact(conversation_id, current_fact, fact_entity_value)
+
+        # Set current_fact to new_fact (returned from ML service)
+        conversation.current_fact = new_fact
+        # db.session.commit()
```
arekmano: remove

src/nlp_service/controllers/nlpController.py
```diff
@@ -0,0 +1,116 @@
...
+        determined_intent = classify_dict['intent']
+        print("Confidence: {}%".format(round(determined_intent['confidence'], 3) * 100))
+        print("Intent: {}".format(determined_intent['name']))
+        return determined_intent['name']
+    else:
+        return None
+
+
+# Determine confidence of returned intent
+def __is_answer_sufficient(classify_dict):
```
arekmano: wouldn't this belong in the RasaClassifier?

src/nlp_service/feature_extraction/Models/compliment.py
```diff
@@ -1,4 +1,5 @@
 from feature_extraction.GlobalVariables import regex
+
 from feature_extraction.Models import Abstract_Model
```
arekmano: Are we still using these files in our classification?

src/nlp_service/gram_classify/introduction_parser.py
```diff
@@ -1,4 +1,4 @@
-from models.introduction import IntroductionOutput
+from models.classifier.introduction import IntroductionOutput
```
arekmano: we can rm this file

arekmano (changes requested):

.gitignore
```diff
@@ -4,4 +4,5 @@
 #PyCharm
 .idea/
 src/__pycache__/
```
arekmano: you can remove this. same as below

src/nlp_service/controllers/nlpController.py
TaimoorRana: 👍

src/nlp_service/services/mlService.py
```diff
@@ -0,0 +1,26 @@
+import requests
+
+ML_URL = "http://ml_service:3001"
+
+
+# This is a placeholder until the interface for the ML service is defined.
+# Returns first fact to ask
```
TaimoorRana: Please use docstrings

src/postgresql_db/models.py
```diff
@@ -40,54 +52,86 @@ class DocumentType(Enum):
 '''
 
 
+class Fact(db.Model):
+    id = db.Column(db.Integer, primary_key=True)
+
+    # Attributes
+    name = db.Column(db.String(50), nullable=False)
```
TaimoorRana, line 61: Consider assigning string length to a variable since it’s being used a few times in this Class

src/nlp_service/services/mlService.py
ghost: Noted

Vynny:
Everything is working but RasaClassifier is causing the docker container to run out of memory.
Investigating this currently.

src/nlp_service/feature_extraction/Models/compliment.py
Samuel-Campbell: no. please delete this

src/nlp_service/feature_extraction/Models/proposition.py
```diff
@@ -1,4 +1,5 @@
-from feature_extraction.Models import clause as c, compliment as cmp, predicate as p
```
Samuel-Campbell: delete this file

src/nlp_service/feature_extraction/Models/proposition.py
```diff
@@ -1,4 +1,5 @@
...
+from feature_extraction.Models import clause as c
+from feature_extraction.Models import compliment as cmp, predicate as p
```
Samuel-Campbell: delete this file

src/nlp_service/feature_extraction/fact_extraction.py
```diff
@@ -1,6 +1,7 @@
 import re
-from feature_extraction.WordToInteger.word_to_int import WordToInt
+
 from feature_extraction.FactTree.PropositionLogic import Proposition
+from feature_extraction.WordToInteger.word_to_int import WordToInt
```
Samuel-Campbell: delete file

src/nlp_service/rasa/app_test.py
```diff
@@ -0,0 +1,26 @@
+import os
```
Samuel-Campbell: missing assertions

src/nlp_service/feature_extraction/fact_extraction.py
Vynny: We'll just delete the entire module. If we need it later, we'll fish it out.

src/backend_service/controllers/conversationController.py
```diff
@@ -149,7 +153,7 @@ def upload_file(conversation_id, file):
 ##################
 
 def __get_conversation(conversation_id):
```
arekmano: we need some documentation on this method. The if/elif statements need explaining

src/nlp_service/gram_classify/classifiers/has_lease_expired_classifier.py
```diff
@@ -1,7 +1,8 @@
 # -*- coding: utf-8 -*-
-from services.gram_classifier import GramClassifier
 import os
 
+from gram_classify.gram_classifier import GramClassifier
+
 
 class HasLeaseExpiredClassifier(GramClassifier):
```
arekmano: are we still using these classifiers?

src/backend_service/controllers/conversationController.py
Vynny: __get_conversation()?
I don't think it needs documentation. If the conversation exists it's returned, else it aborts with a 404.

src/backend_service/app_test.py
```diff
@@ -14,25 +16,8 @@
 ################
```
TaimoorRana: Please use docstrings

Vynny: That isn't documenting anything, it's just marking sections in the test file since its not using unittest at the moment.

src/nlp_service/app.py
```diff
@@ -1,59 +1,28 @@
...
+
+
+@app.route("/claim_category", methods=['POST'])
+def classify_claim_category():
+    input = request.get_json()
+    return nlpController.classify_claim_category(input['conversation_id'], input['message'])
+
+
+@app.route("/submit_message", methods=['POST'])
+def submit_message():
```
TaimoorRana, line 32: what does this return? from the name, it seems like a void method but seems to be returning something

src/nlp_service/controllers/nlpController.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def __classify_claim_category(message):
+    classify_dict = rasaClassifier.classify_problem_category(message)
+    print(classify_dict)
+
+    # Return the claim category, or None if the answer was insufficient in determining one
+    if __is_answer_sufficient(classify_dict):
+        determined_problem_category = classify_dict['intent']
+        print("Confidence: {}%".format(round(determined_problem_category['confidence'], 3) * 100))
```
TaimoorRana: are these being printed for debugging purposes?

src/nlp_service/rasa/rasa_classifier.py
```diff
@@ -0,0 +1,117 @@
...
+
+    @staticmethod
+    def intent_percent_difference(intent_dict):
+        intent_ranking = intent_dict['intent_ranking']
+
+        confidence_top = intent_ranking[0]['confidence']
+        confidence_contender = intent_ranking[1]['confidence']
+
+        percent_difference = abs(confidence_contender - confidence_top) / (
+            0.5 * (confidence_contender + confidence_top))
```
TaimoorRana, line 115: why 0.5? will this value ever be changed in the future?

src/nlp_service/app.py
Vynny: The documentation for the return is in nlpController.
These are simply the HTTP routes, all the important functionality is in nlpController. This is done the same way in backend_service so that app.py is not polluted.

src/postgresql_db/models.py
```diff
@@ -40,64 +54,128 @@ class DocumentType(Enum):
...
 
 
 class File(db.Model):
     id = db.Column(db.Integer, primary_key=True)
+
+    # Foreign Keys
     conversation_id = db.Column(db.Integer, db.ForeignKey('conversation.id'))
+
+    # Attributes
     name = db.Column(db.String(50), nullable=False)
```
Taimoorrana1: please assign these values to variables

src/nlp_service/controllers/nlpController.py
Vynny: These would go into logs, however since we don't have logging functionality at the moment they're just being printed out to the console.

src/nlp_service/rasa/rasa_classifier.py
Vynny: It's part of the formula:

Percent difference of x and y
abs(x-y) / (1/2 * (x + y))

TaimoorRana (changes requested):
See the comments I left

src/nlp_service/controllers/nlpController.py
TaimoorRana: there is a logger in /src/ml_service/outputs/output.py maybe you can use that

src/backend_service/controllers/conversationController.py
mihaiqc: Documentation for a getter?

src/postgresql_db/models.py
ghost: Why? It looks odd and out of place when doing so. We have string length 50, 40, 255, 100. 50 is used 3 times. There's absolutely no reason to do so. If I do one for String(50), I need one for the others. If I do one for all of them, it looks unnecessary.

src/nlp_service/controllers/nlpController.py
ghost: Print allows us to go much faster during development as quick changes and tweaks are necessary during such refactoring measures taken during this PR.

Maybe in the future output.py will be used, as of now, it was not necessary.

src/postgresql_db/models.py
mihaiqc: This is unnecessary and even cumbersome to us.

src/nlp_service/controllers/nlpController.py
Vynny: We shouldn't have an ad-hoc logger.
We should eventually add an actual logging solution with log levels. 
We can discuss it in the next meeting, but we'll remove the prints.

src/postgresql_db/models.py
Taimoorrana1: why is that?

src/postgresql_db/models.py
TaimoorRana: I find it's better because if you need to change 50 to 60, then you have to change that in several places in this class. It's better if we had to only change it in 1 location.

src/backend_service/controllers/conversationController.py
ghost: Fixed.

src/backend_service/app_test.py
ghost: Added Documentation for tests.

src/nlp_service/app.py
ghost: Added documentation for routing mechanism

src/postgresql_db/models.py
ghost: This is 3-4 times less intuitive and more confusing than just delimiting the size of the column with a [simple number](http://flask-sqlalchemy.pocoo.org/2.3/models/).

Samuel-Campbell (approved):
Guys it looks fine. we can refactor docstring or wtv later. it's not important

ghost (approved):

TaimoorRana (approved):

src/backend_service/app.py
```diff
@@ -1,21 +1,23 @@
+import os
+
 from flask import Flask, request, abort, make_response, jsonify
 from flask_cors import CORS
-from flask_marshmallow import Marshmallow
 
-import database
-import os
+import util
+from controllers import conversationController, legalController
```
arekmano, line 7: use snake_scase. keep style consistent (NOT a blocker)

src/nlp_service/controllers/nlpController.py
```diff
@@ -0,0 +1,178 @@
...
+
+# Global Variables
+
+# Decided value of 30% percent difference which is used to determine if a clarification is needed.
+# Based on percent difference of two highest intents determined.
+minimum_percent_difference = 0.3
+
+# Rasa Classifier - Initialization of RasaClassifier, used for claim category determination and fact value classification
+rasaClassifier = RasaClassifier()
+rasaClassifier.train(force_train=True)
```
arekmano: this will slow boot time significantly as we expand RASA (non-blocking comment ...for now)

src/nlp_service/controllers/nlpController.py
```diff
@@ -0,0 +1,178 @@
...
+    # Save first fact as current fact
+    conversation.current_fact = first_fact
+
+    # Commit
+    db.session.commit()
+
+    # Generate next message
+    first_fact_question = Responses.fact_question(first_fact.name)
+    message = Responses.chooseFrom(Responses.category_acknowledge).format(
+        claim_category=conversation.claim_category.value.lower().replace("_", " "),
```
arekmano, line 70: why are we doing this replace magic?

src/nlp_service/rasa/rasa_classifier.py
arekmano: This is a bug waiting to happen IMO.

This works well when there are only 2 intents that matter. Consider 0.3 0.2 0.1 0.1 etc... This percent difference will accept 0.3 as a high enough confidence, Whereas it is only 30% sure that this is the correct category, and it should cause reprompt since the classifier is not sure.

This is a borderline blocker for me. This heuristic cannot be the best way of doing things, especially when thresholding can be done. Either put a TODO comment here to address in the future, when we classify multiple intents or we find a solution before PR goes through

arekmano (changes requested):
See Percent difference comment. After that fix, you can dismiss my review/ping me so I approve it.

src/nlp_service/controllers/nlpController.py
Vynny: Yeah I definitely need to fix this. I set it to true because even without force train, rasa seems to take forever loading pre trained data. It takes 1 second less to load pre trained models than to just retrain, so I might be missing something from the docs.

I'll add a fix for this in another PR. Right now it's fast enough so that it doesn't had adverse effects.

src/nlp_service/controllers/nlpController.py
Vynny: This could be made a bit nicer, but it's essentially taking the value of the claim category enum and removing the _.
LEASE_TERMINATION -> lease termination -> I see you're having problems with lease

src/nlp_service/rasa/rasa_classifier.py
Vynny: Good point. This needs some research put into it by the NLP team.
It was done this way because right now our facts only do have 2 intents that matter (boolean or text with 2 possible values).
I'll add a 50% threshold and a todo indicating it should be looked into thoroughly for the next iteration.

arekmano (approved):

src/nlp_service/rasa/rasa_classifier.py
ghost: That's exactly the issue but with the threshold. It will be absolutely impossible to guarantee a set threshold over every model unless we specify a threshold per model due to multiple intents. Unless we all go one-by-one and polish the data to an absurd extent.

Ex: 
Model A happens to have generated 100 intents.
Model B happens to have generated 3 intents.
Threshold set at 80%.

Model B can generated Intent1, Intent2, Intent3 respectively 86, 16, 8. Which, in this case is feasible.
Model A though, 40,1, 1, 1, 1^(95).

Model A will fail even though it is clearly a lot better than the sum of the rest of the confidence percentages.

src/nlp_service/controllers/nlpController.py
```diff
@@ -177,6 +177,6 @@ def __is_answer_sufficient(classify_dict):
     if len(classify_dict['intent_ranking']) > 1:
         percent_difference = RasaClassifier.intent_percent_difference(classify_dict)
         highest_intent_confidence = classify_dict['intent']['confidence']
-        if highest_intent_confidence < minimum_intent_confidence_threshold and percent_difference < minimum_percent_difference:
+        if highest_intent_confidence < minimum_intent_confidence_threshold or percent_difference < minimum_percent_difference:
```
arekmano: should be and

arekmano (changes requested):

arekmano (approved):

PR #149, opened 2017-11-10
[#75] Allow any service to import  modules from the database service

lancelafontaine:
[#75]

- [x]  Restructures application to import modules from the database service for the backend and NLP services.


Vynny (approved):
lgtm

Issue #156, opened 2017-11-11
Dev Story: Address Technical Debt


lancelafontaine:
*Description*
As a developer, I would like to remove as many sources of technical debt as possible

Our development speed has now slowed down due to the technical debt that we've accumulated throughout the first release. Travis CI build times are > 20 minutes, legacy code exists within our system that is unused and local development is made more difficult because of the tight coupling between all our micro-services.  With the first release coming to an end, it is time to remove some of the technical debt that cannot be addressed within a specific user story.

*Scope of Work*
- [x] Remove all unused dependencies from the `requirements.txt`/`package.json` files 
- [x] ~Restructure `requirements.txt` files to be composable and usable for different purposes~ Deemed unnecessary. Splitting into `requirements.txt` and `requirements_test.txt` is good enough for now.
- [x] Write scripts, Dockerfile, or detailed README for installation of dependencies
- [x] ~Reduce build time by caching Docker images (without ML models) on Docker Hub~. This is not a good solution, since will need to download the ML models either way, and caching would be extremely minimal. 
- [x] Minimize Docker image size by cleaning up installation artifacts or unused dependencies. 
- [x] Address concerns over build contexts and sharing modules in microservice architecture
- [x] Continuous delivery is unreliable

*Acceptance Criteria / Demo Requirements*
- All processes are scripted or heavily documented 
- Docker build times are significantly shorter.

**Documentation, files, diagrams, etc. relating to this feature:**
...

***Changes***:
- [Optional]

arekmano:
This story is being added as technical debt is slowing down our velocity a noticeable amount.

naregeff:
they managed to convince me of the necessity. signed off.

PR #168, opened 2017-11-12
[#75] Adds another database state management service to apply migrations in the future.

lancelafontaine:
[#75] 

- [x] Adds another database state management service to apply migrations in the future.


src/nlp_service/services/mlService.py
```diff
@@ -1,6 +1,6 @@
 import requests
 
-from postgresql_db.models import Fact, FactEntity
+from db_state_service.models import Fact, FactEntity
```
arekmano: I personally prefer duplicating the models' code rather than breaking microservice architecture

lancelafontaine: This is something we should address as scope of work for issue #156 in iteration 5.

arekmano (approved):
LGTM

lancelafontaine:
To be reopened again  in the future as part of issue #156

Issue #65, opened 2017-10-13
Reprompt on ambiguous answer


arekmano:
*Description*
As a user, I'd like the system to ask me for further precision regarding the questioning phase of my case if my answer was unclear or ambiguous.

*Scope of Work*
- [x] Determine some measures to implement a confidence degree in the answer generation process
- [x] Set a percentage of acceptable confidence ratio through trial and error
- [x] If the answer of the user is too ambiguous ( |Most Sure - 2nd Most Sure| / [1/2(Most Sure + 2nd Most Sure)] ), display a message asking the user to confirm the understanding of the system.
- [x] Create a High-fidelity mock-up of interaction (User inputs his/her problem, system needs clarification due to too low % difference)
- [x] Implement an outlier detection pre-processing step in order to filter out garbage input.

*Story Points*
- 8

*Priority*
- High

*Risk*
- High

*Demo requirement*
1. The user attempts to trick the system with a misspelling of an important noun or verb relating to a fact i.e. "I am a tanant" 
2. The system asks the user to repeat the sentence i.e. "Please repeat" or "Did you mean that you are a tenant?"

*Acceptance Criteria*
- The system is able to ask users for confirmation to ambiguity of facts or messages
- The users sees the confirmation message
- The user can confirm the message or deny and rewrites a clearer message (with the possibility of the first two steps repeating themselves)

naregeff:
Verified, high priority.

Samuel-Campbell:
'OTHER' classifier
---------------------------------------------
Create a classifier of type 'other' which will trigger the ambiguous answer response.
Train this vector on many statements not pertaining to any supported topics.

Samuel-Campbell:
I suggest making an 'other' category. I'm not sure if RASA takes care of it already though.

arekmano:
I'm unsure. However, this will most likely need to involve some form of outlier detection, before the phrase is classified. I suggest splitting this story into 2. one for the UI and another for the outlier/anomaly detection method.

Some Reading links:
https://medium.com/towards-data-science/a-brief-overview-of-outlier-detection-techniques-1e0b2c19e561
https://en.wikipedia.org/wiki/Anomaly_detection

ghost:
This is being heavily worked on in junction with #75

mihaiqc:
![image](https://user-images.githubusercontent.com/9852083/32695136-63858222-c721-11e7-9d6e-242c351b470a.png)

High Fidelity Model.
[...]

PR #170, opened 2017-11-13
[#65] Fixes Travis CI build failure reporting ASAP

lancelafontaine:
[#65, #75, #105]

- [x] Fixes Travis CI build failure reporting
- [x] Begins work for fixing backend unit tests and database mocking
- [x] Begins work for correcting Python code coverage reporting (should not be reporting Python test files as untested)

Since multiple services are currently failing their unit tests, we will merge this fix in even though the build is not passing, just so that:
- multiple subteams have the chance to fix their unit tests in their own branches.
- master should theoretically be failing because of unit tests at the moment, even though it is not.

### DIFF
--- a/src/backend_service/Dockerfile
+++ b/src/backend_service/Dockerfile
@@ -6,8 +6,8 @@ RUN mkdir -p /usr/src/app/src/backend_service
 
 COPY ./src/backend_service/requirements.txt ./src/backend_service/
 COPY ./src/backend_service/requirements_test.txt ./src/backend_service/
-RUN pip install -r ./src/backend_service/requirements_test.txt
-
+RUN pip install http://initd.org/psycopg/upload/psycopg2-2.7.3.1.dev0/psycopg2-2.7.3.1.dev0-cp36-cp36m-manylinux1_x86_64.whl && \
+    pip install -r ./src/backend_service/requirements_test.txt
 COPY . .
 
 WORKDIR /usr/src/app/src/backend_service
--- a/src/postgresql_db/database.py
+++ b/src/postgresql_db/database.py
@@ -1,12 +1,14 @@
 from flask_sqlalchemy import SQLAlchemy
+import os
 
 
 def connect(app, user, password, db, host='postgresql_db'):
-    # We connect with the help of the PostgreSQL URL
-    url = 'postgresql://{user}:{password}@{host}/{db}'.format(user=user, password=password, host=host, db=db)
-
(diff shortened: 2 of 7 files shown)


Vynny (approved):

src/backend_service/Dockerfile
```diff
@@ -6,8 +6,8 @@ RUN mkdir -p /usr/src/app/src/backend_service
 
 COPY ./src/backend_service/requirements.txt ./src/backend_service/
 COPY ./src/backend_service/requirements_test.txt ./src/backend_service/
-RUN pip install -r ./src/backend_service/requirements_test.txt
-
+RUN pip install http://initd.org/psycopg/upload/psycopg2-2.7.3.1.dev0/psycopg2-2.7.3.1.dev0-cp36-cp36m-manylinux1_x86_64.whl && \
```
Vynny, line 9: What's this for?

src/postgresql_db/database.py
```diff
@@ -1,12 +1,14 @@
...
-    db = SQLAlchemy(app)
-
+    if os.environ['CI']:
+        app.config['SQLALCHEMY_DATABASE_URI'] = 'sqlite:///app.db'
+        return SQLAlchemy(app)
+    else:
+        # We connect with the help of the PostgreSQL URL
+        url = 'postgresql://{user}:{password}@{host}/{db}'.format(user=user, password=password, host=host, db=db)
+        app.config['SQLALCHEMY_DATABASE_URI'] = url
+        db = SQLAlchemy(app)
```
Vynny, line 13: can be simplified to return SQLAlchemy(app). Do the app configuration then return the object.

src/backend_service/Dockerfile
lancelafontaine: That `psycopg2` seemed to be suddenly failing to build properly, because of an issue with the binary in wheel files. I found this fix here: https://github.com/psycopg/psycopg2-wheels/issues/2#issuecomment-324898857

This should hopefully be a temporary fix. I'll address reintegrating back into `requirements.txt` in a future PR.

src/postgresql_db/database.py
lancelafontaine: Saw that, and applied for first condition. It's a very minor style issue though, so I'll address this in a future PR if you don't mind (just to address the more pressing issue of tests failing now)

lancelafontaine:
Force-merging now (please read description for rationale behind this).

Issue #151, opened 2017-11-10
DEV STORY: Create Machine-Readable Representations of Precedents


arekmano:
*Description*
As a dev, I would like a vector representation of precedents so it can be used for ML.

*Scope of Work*
- [x] Create a BINARY file representing Facts and Outcomes for each precedent

*Acceptance Criteria*
- Binary file can be used to train desired models

arekmano:
Tentative file layout:

> {
  "precedent": "AZ-511123",
  "facts": [1,0,1,0,1,0,1,0...],
  "outcomes": [0,1,0,0,0,1...],
  "lease_termination": 0
}

Samuel-Campbell:
Depends on part of #153

naregeff:
they managed to explain it to me. good stuff. signed off.

PR #188, opened 2017-11-16
[#151] Precedents data structure

TaimoorRana:
[#151]

- Created structure for precedent data
- This data can also be outputted as JSON

**Precedent data structure:**
```
{
"AZ-1234561":{
              "facts_vector": [0,1,1,1,1,1,0,0,0,1,0]
              "decisions_vector": [0,1,1,1,1,1,0,0,0,1,0]
            }
"AZ-1234567":{
              "facts_vector": [0,1,1,1,1,1,0,0,0,1,0]
              "decisions_vector": [0,1,1,1,1,1,0,0,0,1,0]
             }
}
```


Samuel-Campbell:
I disagree. 

"AZ-123456":{
              "facts": [(0,"some fact"),(10,"another fact with different category")...] -> Don't need
              "facts_vector": [0,1,1,1,1,1,0,0,0,1,0] -> not binary
              "outcomes": [(0,"some fact"),(10,"another fact with different category")...] -> Don't need
              "outcomes_vector": [0,1,1,1,1,1,0,0,0,1,0] -> Not binary
              "lease_terminated": None --> don't need. It's represented in the outcomes
            }

src/ml_service/precedents/structured_precedents.py
```diff
@@ -0,0 +1,61 @@
...
+class StructuredPrecedent:
+
+    CATEGORY_INDEX = 0
+    FACTS = "facts"
+    FACTS_VECTOR = "facts_vector"
+    OUTCOMES = "outcomes"
+    OUTCOMES_VECTOR = "outcomes_vector"
+    LEASE_TERMINATED = "lease_terminated"
+
+    def __init__(self, fact_labels, fact_data_tuple, outcome_labels, outcome_data_tuple):
```
Samuel-Campbell: It seems to be like you want to load the precedent binaries. Use the Cluster models instead.

Samuel-Campbell (changes requested):

src/ml_service/precedents/structured_precedents.py
```diff
@@ -0,0 +1,61 @@
...
+                if data_type == self.FACTS:
+                    self.precedents[file_name][self.FACTS_VECTOR][raw_fact_tuple[self.CATEGORY_INDEX]] = 1  # 1 signifies that the fact exists
+                    self.precedents[file_name][self.FACTS].append(raw_fact_tuple)
+                else:
+                    self.precedents[file_name][self.OUTCOMES_VECTOR][raw_fact_tuple[self.CATEGORY_INDEX]] = 1  # 1 signifies that the fact exists
+                    self.precedents[file_name][self.OUTCOMES].append(raw_fact_tuple)
+
+    def write_data_to_output_dir(self):
+        with open("precedent_JSON.txt", 'w') as output_file:
+            json.dump(self.precedents, output_file)
```
Samuel-Campbell: Add unittesting for your .py file

Samuel-Campbell (changes requested):

src/ml_service/precedents/structured_precedents.py
```diff
@@ -0,0 +1,61 @@
...
+                                                       enumerate(data_tuple[InformationType.PRECEDENTS_FILE_NAMES.value][labels == label])):
+
+                file_name = file_name_tuple[1].replace(".txt", "")
+
+                if file_name not in self.precedents:
+
+                    # Create new entry with file name
+                    self.precedents[file_name] = dict.fromkeys([self.FACTS, self.FACTS_VECTOR, self.OUTCOMES, self.OUTCOMES_VECTOR, self.LEASE_TERMINATED])
+
+                    # initialize fields
```
Samuel-Campbell: Do not use lists. Use numpy arrays. it will significantly reduce the training time

Samuel-Campbell (changes requested):

src/ml_service/precedents/structured_precedents.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def create_vector(self, labels, data_tuple, data_type):
+        """
+        Creates a fact or outcome vector for all precedents
+        :param labels (int): Cluster labels
+        :param data_tuple ([int], [string], [string]): vectors, transformed sentences, original sentence
+        :param data_type (string): "facts" or "outcomes"
+        :return: Void
+        """
+        unique_labels = set(labels)
```
Samuel-Campbell: the labels are only integers from -1 to n with no significance. 
set(labels) won't do anything because all labels are unique

src/ml_service/precedents/structured_precedents.py
TaimoorRana: yes, but I also need the cluster labels and for that, some X clusters need to run. So this way, the class can be used by any Clusters. Just the precedent binaries aren't enough. What do you think?

src/ml_service/precedents/structured_precedents.py
TaimoorRana: Will do

arekmano:
Save as pickle too please. More efficient to load

arekmano (changes requested):
Pickle

TaimoorRana:
@Samuel-Campbell please see #151 for _"lease_terminated": None --> don't need. It's represented in the outcomes_

src/ml_service/precedents/structured_precedents.py
TaimoorRana: Yes but I don't want repetitions

src/ml_service/precedents/structured_precedents.py
Samuel-Campbell: I don't understand what you are trying to say. what are you trying to do exactly? bullet pts

Samuel-Campbell:
@Taimoorrana1 it's only a suggested template. We don't actually need that value because we should know it's index in the array

src/ml_service/precedents/structured_precedents.py
Taimoorrana1: this need clusters labels, which we get after a cluster runs

TaimoorRana: I need the category (labels) for each fact and outcome to create the vectors. These labels are produced after a cluster has run.

arekmano:
Though it may correspond to multiple clusters I think it's correct like this

src/ml_service/precedents/structured_precedents.py
Samuel-Campbell: ohhh ok I see what you're doing now

src/ml_service/precedents/structured_precedents.py
Samuel-Campbell: It's ok i see what you're doing. I had a hard time understanding at first

src/ml_service/precedents/structured_precedents.py
```diff
@@ -0,0 +1,61 @@
+from global_variables.global_variable import InformationType
+import json
+
+
+class StructuredPrecedent:
+
+    CATEGORY_INDEX = 0
```
dodoels: i think we should formalize what we use for naming to avoid confusion. `category`? `label`? `feature`?

src/ml_service/precedents/structured_precedents.py
```diff
@@ -0,0 +1,61 @@
...
+                    # Create new entry with file name
+                    self.precedents[file_name] = dict.fromkeys([self.FACTS, self.FACTS_VECTOR, self.OUTCOMES, self.OUTCOMES_VECTOR, self.LEASE_TERMINATED])
+
+                    # initialize fields
+                    self.precedents[file_name][self.FACTS_VECTOR] = [0] * len(unique_labels)
+                    self.precedents[file_name][self.FACTS] = []
+                    self.precedents[file_name][self.OUTCOMES_VECTOR] = [0] * len(unique_labels)
+                    self.precedents[file_name][self.OUTCOMES] = []
+
+                # populate fields
```
dodoels: Do we need this? I think the vectors are already referring the facts in the clusters.

TaimoorRana: which vectors?

arekmano (approved):
LGTM

dodoels (approved):

Samuel-Campbell (approved):

src/ml_service/precedents/structured_precedents.py
```diff
@@ -0,0 +1,122 @@
...
+import joblib
+import os
+from global_variables.global_variable import InformationType
+from global_variables.global_variable import Global
+from ml_models.models import Load
+
+
+class StructuredPrecedent:
+
+    FACTS = "facts"
```
arekmano, line 11: not needed. 1s and 0s are all I need

src/ml_service/precedents/structured_precedents.py
```diff
@@ -0,0 +1,122 @@
...
+    FACTS_VECTOR = "facts_vector"
+    DECISIONS = "decisions"
```
arekmano, line 13: not needed. 1s and 0s are all I need

src/ml_service/precedents/structured_precedents.py
```diff
@@ -0,0 +1,122 @@
...
+            for file_name in data_tuple[InformationType.PRECEDENTS_FILE_NAMES.value][index]:
+                file_name = file_name.replace(".txt", "")
+                if file_name not in self.precedents:
+
+                    # Create new entry with file name
+                    self.precedents[file_name] = dict.fromkeys([self.FACTS, self.FACTS_VECTOR, self.DECISIONS, self.DECISIONS_VECTOR])
+
+                    # initialize fields
+                    self.precedents[file_name][self.FACTS_VECTOR] = numpy.zeros(unique_labels_size, dtype=numpy.int)
+                    self.precedents[file_name][self.FACTS] = []
```
arekmano: don't need these, keep binary small

src/ml_service/precedents/structured_precedents.py
```diff
@@ -0,0 +1,122 @@
...
+                    self.precedents[file_name][self.DECISIONS_VECTOR] = numpy.zeros(unique_labels_size, dtype=numpy.int)
+                    self.precedents[file_name][self.DECISIONS] = []
```
arekmano: don't need these, keep binary small

src/ml_service/precedents/structured_precedents.py
TaimoorRana: removed

src/ml_service/precedents/structured_precedents.py
TaimoorRana: removed

PR #189, opened 2017-11-16
[#156] Optimizes existing Docker builds

lancelafontaine:
[#156]

- [x] Removes all volumes in development. This was proving to be troublesome because it would write new files to the host filesystem on every `./cjl build`, which means that nothing ever got cached.
- [x] Creates a global `.dockerignore` file to reduce number of files sent to the Docker daemon as part of the build context.
- [x] Performs heavier caching and layering for `./cjl build` for every service

Previous builds on the local machine would take approximately 5 or 10 minutes. Here are some very unofficial benchmarks that these changes bring in specific instances:

- If no files were changed since your last build: `4.722s`
- If you changed a "trivial" source file (almost everything except `requirements.txt`/`package.json` and `init.py`): `38.093s`

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,4 @@
+comment:
+  behavior: default
+  layout: header, diff
+  require_changes: false
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,4 +1,0 @@
-comment:
-  behavior: default
-  layout: header, diff
-  require_changes: false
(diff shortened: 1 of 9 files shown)


Vynny (approved):

dodoels (approved):

arekmano (approved):
copy as late as possible. got it :+1:

Issue #183, opened 2017-11-15
Claim: Lease Termination


arekmano:
*Description*
As a user, I would like the system to predict whether or not lease termination is justified

*Scope of Work*
- [x] Create supervised learning algorithm benchmark for evaluating predictions
- [x] Implement supervised learning algorithm that can evaluate whether or not lease termination is justified based on facts present in the case .
- [x] Extract using regex, or similar technique whether or not the lease was terminated in a specific case.  

*Demo requirement*
1. The user selects the facts that are relevant to their case
2. The user sends the facts to the system
3. The system displays "Lease termination is justified" or "Lease termination is not justified" 

*Acceptance Criteria*
- The user is able perform demo steps
- The F1 score of the supervised learning algorithm is >50%.

**Documentation, files, diagrams, etc. relating to this feature:**
Url, books, etc.

***Changes***:
- [Optional]

arekmano:
This issue is blocked by #151

dodoels:
Is it possible to test the model with fake data now? (since #151 will generate a 45000 x 4800 matrix with just 0 and 1 as values) so it doesnt block the implementation

naregeff:
it wow-ed. good stuff. signed off.

arekmano:
Current fact weights, as related to lease termination output:
```
absent : 0.000000000000000000e+00
apartment_impropre : -1.100925369199277082e-01
apartment_infestation : 0.000000000000000000e+00
asker_is_landlord : -1.779103941858892757e-01
asker_is_tenant : -1.368796370477016744e-01
bothers_others : 0.000000000000000000e+00
disrespect_previous_judgement : 0.000000000000000000e+00
incorrect_facts : 0.000000000000000000e+00
landlord_inspector_fees : 0.000000000000000000e+00
landlord_notifies_tenant_retake_apartment : 0.000000000000000000e+00
landlord_pays_indemnity : 0.000000000000000000e+00
landlord_prejudice_justified : 1.783754245445596798e-01
landlord_relocation_indemnity_fees : -4.040827592879288332e-02
landlord_rent_change : -2.180868817701108742e-01
landlord_rent_change_doc_renseignements : 0.000000000000000000e+00
landlord_rent_change_piece_justification : 0.000000000000000000e+00
landlord_rent_change_receipts : 0.000000000000000000e+00
landlord_retakes_apartment : 3.160223470087041298e-01
landlord_retakes_apartment_indemnity : 0.000000000000000000e+00
[...]

PR #190, opened 2017-11-17
[#183] Created Neural Network Lease termination Predictor

arekmano:
- Fixed minor variable name bug
- Added basic 3-layer neural network implementation

[#183]


src/ml_service/outcome_predictor/basic_neural_net.py
```diff
@@ -0,0 +1,53 @@
...
+        vals = [precedent for precedent in self.dictionary.values() if len(precedent['facts_vector']) == 3486]
+
+        # Prep data
+        print('splitting data')
+        x_total = np.array(
+            [np.reshape(precedent['facts_vector'], (3486,)) for precedent in vals])
+        y_total = np.array(
+            [np.reshape(precedent['decisions_vector'], (1,)) for precedent in vals])
+        y_total = to_categorical(y_total)
+        X_train, X_test, y_train, y_test = train_test_split(
```
TaimoorRana: x lowercase please

src/ml_service/train.py
```diff
@@ -0,0 +1,28 @@
...
+    file.close()
+    resilie = 0
+    for (key, val) in model.items():
+        with open(Global.precedent_directory + key + '.txt', 'r', encoding="ISO-8859-1") as f:
+            isFound = False
+            val['decisions_vector'] = np.array([0])
+            for line in f:
+                if "RÉSILIE" in line and isFound is False:
+                    val['decisions_vector'] = np.array([1])
+                    resilie += 1
```
TaimoorRana: Where is this being used?

src/ml_service/outcome_predictor/basic_neural_net.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def train(self):
+        """
+            Trains the neural network.
+            Current config: input vector shape (Vector size: 3486) ->
+                            intermediate vector  (Vector size: 64) ->
+                            output vector (Vector size: 2 (isResiliated, isNotResiliated))
+        """
+        vals = self.dictionary.values()
+        vals = [precedent for precedent in self.dictionary.values() if len(precedent['facts_vector']) == 3486]
```
TaimoorRana: I updated the code in #151, if there's no fact vector then `precedent['facts_vector']` will be `None`. So we can avoid hardcoded values

src/ml_service/train.py
arekmano: it's not. will rm

src/ml_service/outcome_predictor/basic_neural_net.py
arekmano: It is common practice to capitalize matrices. All examples that I learned from followed this rule. Unsure if we want to keep working with this styling or not.

https://datascience.stackexchange.com/questions/17598/why-are-variables-of-train-and-test-data-defined-using-the-capital-letter-in-py#answer-17600

TaimoorRana: good to know why it's done. My preference would be that we stay consistent in our code base.

src/ml_service/outcome_predictor/basic_neural_net.py
```diff
@@ -0,0 +1,54 @@
...
+                            output vector (Vector size: 2 (isResiliated, isNotResiliated))
+        """
+        vals = self.dictionary.values()
+        vals = [precedent for precedent in self.dictionary.values() if precedent[
+            'facts_vector'] is not None]
+
+        # Prep data
+        print('splitting data')
+        x_total = np.array(
+            [np.reshape(precedent['facts_vector'], (3486,)) for precedent in vals])
```
TaimoorRana: instead of 3486, could we `len(precedent['facts_vector'])`.

arekmano:
Fixed. please re-eval

src/ml_service/train.py
```diff
@@ -0,0 +1,34 @@
...
+    model = joblib.load(file)
+    file.close()
+    resilie = 0
+    for (key, val) in model.items():
+        with open(Global.precedent_directory + key + '.txt', 'r', encoding="ISO-8859-1") as f:
+            isFound = False
+            val['decisions_vector'] = np.array([0])
+            del val['decisions']
+            del val['facts']
+            for line in f:
```
TaimoorRana: Can you replace f with file. It’s more meaningful. Not a blocker, just me being picky.

src/ml_service/outcome_predictor/basic_neural_net.py
```diff
@@ -0,0 +1,56 @@
...
+
+class BasicNeuralNet(object):
+
+    def __init__(self, dictionary):
+        self.dictionary = dictionary
+
+    def train(self):
+        """
+            Trains the neural network.
+            Current config: input vector shape (Vector size: 3486) ->
```
TaimoorRana: Need to update comments

src/ml_service/outcome_predictor/basic_neural_net.py
```diff
@@ -0,0 +1,54 @@
...
+        vals = [precedent for precedent in self.dictionary.values() if precedent[
+            'facts_vector'] is not None]
+
+        # Prep data
+        print('splitting data')
+        x_total = np.array(
+            [np.reshape(precedent['facts_vector'], (len(precedent['facts_vector']),)) for precedent in vals])
+        y_total = np.array(
+            [np.reshape(precedent['decisions_vector'], (len(precedent['decisions_vector']),)) for precedent in vals])
+        y_total = to_categorical(y_total)
```
dodoels: `x_total` doesn't need to be converted to binary matrix?

src/ml_service/train.py
```diff
@@ -0,0 +1,34 @@
...
+def load_data(useCached=False):
+    if useCached:
+        file = open('structured_precedent_updated.bin', 'rb')
+        return joblib.load(file)
+    file = open('structured_precedent.bin', 'rb')
+    model = joblib.load(file)
+    file.close()
+    resilie = 0
+    for (key, val) in model.items():
+        with open(Global.precedent_directory + key + '.txt', 'r', encoding="ISO-8859-1") as f:
```
dodoels: so `with` keyword is for scoping the file stream?

src/ml_service/train.py
```diff
@@ -0,0 +1,34 @@
...
+            isFound = False
+            val['decisions_vector'] = np.array([0])
+            del val['decisions']
+            del val['facts']
+            for line in f:
+                if "RÉSILIE" in line and isFound is False:
+                    val['decisions_vector'] = np.array([1])
+                    isFound = True
+    print('save update')
+    joblib.dump(model, 'structured_precedent_updated.bin')
```
dodoels: `structured_precedent_resilie`, why are we overriding the binary?

src/ml_service/train.py
```diff
@@ -0,0 +1,34 @@
...
+    file.close()
+    resilie = 0
+    for (key, val) in model.items():
+        with open(Global.precedent_directory + key + '.txt', 'r', encoding="ISO-8859-1") as f:
+            isFound = False
+            val['decisions_vector'] = np.array([0])
+            del val['decisions']
+            del val['facts']
+            for line in f:
+                if "RÉSILIE" in line and isFound is False:
```
dodoels: I see you are filtering `resilie` in original precedent text, but

we have precedents raw data -> outcome/fact clusters -> precedent vectors, and now we are dealing with precedent vectors by going back to the raw data, it doesn't seem like a good idea to me. Maybe we just need to look into `val['decisions']` instead and delete it later.

dodoels: also for my part I can regex `resilie` in this block of codes

src/ml_service/train.py
```diff
@@ -0,0 +1,34 @@
...
+import pdb
+import numpy as np
+
+
+def load_data(useCached=False):
+    if useCached:
+        file = open('structured_precedent_updated.bin', 'rb')
+        return joblib.load(file)
+    file = open('structured_precedent.bin', 'rb')
+    model = joblib.load(file)
```
dodoels: `model`? it is confusing. Maybe just use `precedent` or `precedent_data`

src/ml_service/feature_extraction/preprocessing/precedence_parse_test.py
```diff
@@ -38,7 +38,7 @@ def test_parse_files(self):
         file.writelines('[10] Le chat veut me tuer.\n')
         file.close()
 
-        Global.precedence_directory = __full_path
+        Global.precedent_directory = __full_path
```
dodoels, line 41: please global search and change as well 
`precedence_directory` is still in unit test and failing the build

src/ml_service/outcome_predictor/basic_neural_net.py
arekmano: no. to_categorical changes the value to a 1hot encoding. The input values are already 1 hot encoded

https://hackernoon.com/what-is-one-hot-encoding-why-and-when-do-you-have-to-use-it-e3c6186d008f

src/ml_service/train.py
arekmano: yes

src/ml_service/train.py
arekmano: agreed. this was a prototype of the ML, rather than using the clusters

src/ml_service/train.py
arekmano: it took a long time to do this.which is why i cached it. Will improve with new commit

src/ml_service/train.py
arekmano: ok

src/ml_service/feature_extraction/preprocessing/precedence_parse_test.py
arekmano: ok

src/ml_service/outcome_predictor/basic_neural_net.py
```diff
@@ -0,0 +1,52 @@
...
+    def train(self):
+        """
+            Trains the neural network.
+            Current config: Input -> 64 Nodes -> Outputs
+        """
+
+        # import pdb;pdb.set_trace()
+        # Prep data
+        print('splitting data')
+        x_total = np.array(
```
Samuel-Campbell, line 24: I thought the default was to capitalize those letters X, Y. Not blocking just mentioning

src/ml_service/train.py
```diff
@@ -0,0 +1,72 @@
+from global_variables.global_variable import Global
```
Samuel-Campbell, line 1: Do we want this in the root directory? NOT blocking

src/ml_service/train.py
```diff
@@ -0,0 +1,72 @@
...
+            del val['facts']
+        resiliation_values = [val['decisions_vector'][x]
+                              for x in resiliation_custers]
+        if np.sum(resiliation_values) > 0:
+            val['decisions_vector'] = np.array([1])
+        else:
+            val['decisions_vector'] = np.array([0])
+    return valid_values
+
+print('loading data')
```
Samuel-Campbell, line 69: if __name__ == '__main__' . Not a blocker

src/ml_service/outcome_predictor/basic_neural_net.py
```diff
@@ -0,0 +1,52 @@
...
+            Dense(units=len(y_total[0]), activation='softmax'))
+        model.compile(loss='categorical_crossentropy',
+                      optimizer='sgd',
+                      metrics=['accuracy'])
+        # Train
+        print('Starting training')
+        model.fit(x_train, y_train, epochs=5, batch_size=32)
+
+        # Test
+        print('Starting testing')
```
Samuel-Campbell, line 47: can you use the logger found in ml_service/src/output/output.py instead of print statements  -- Not blocker

Samuel-Campbell (commented):
None of my comments are blockers. The logger instead of print would be very beneficial

src/ml_service/outcome_predictor/basic_neural_net.py
Samuel-Campbell: The logger I wrote prints to the terminal and logs to textfile

TaimoorRana:
Looks good

dodoels (approved):

src/ml_service/outcome_predictor/basic_neural_net.py
arekmano: it is. but as taimoor mentioned that our coding style was keeping the variables as lowercase, I changed it

src/ml_service/outcome_predictor/basic_neural_net.py
arekmano: will change in new PR

src/ml_service/train.py
arekmano: will move to subfolder in new PR

src/ml_service/train.py
arekmano: in new PR

PR #195, opened 2017-11-19
[#65] Implements a basic outlier detection system for user messages with RASA training data as input.

lancelafontaine:
[#65] 

- [x] Implements a basic outlier detection system for user messages with RASA training data as input.

Our input data set is quite small, making the outlier detection reject most responses. We will need to either generate lots more sentences ourselves as RASA training data, or eventually base it off user-provided data.


src/nlp_service/controllers/nlpController.py
```diff
@@ -152,6 +158,13 @@ def __classify_claim_category(message):
 
 
 def __extract_entity(current_fact_name, message):
+    # First pass: outlier detection
+    # TODO: For now, this is disabled while we are gathering data from beta users
+    if False:
```
Samuel-Campbell: ?

Samuel-Campbell:
looks good except for that weird if False
Also add comments.
I trust you will make the changes so i approved the PR

Samuel-Campbell (approved):

src/nlp_service/controllers/nlpController.py
lancelafontaine: Outlier detection will be turned off until we amass enough user-generated data for it to be useful.

Until then, we also want all inputs to be accepted (during the beta testing).

Vynny: I think that this should be a config setting when starting the application. However, it can be implemented in another pr.

I don't like this because it essentially renders the application un-demoable unless you change code.

src/nlp_service/outlier/outlier_detection.py
```diff
@@ -0,0 +1,62 @@
...
+        all_sentences = []
+        for fact_path in rasa_fact_paths:
+            with open(fact_path, 'r') as f:
+                file_json = json.loads(f.read().encode('utf-8'))
+            for example in file_json['rasa_nlu_data']['common_examples']:
+                all_sentences.append(example['text'].lower())
+
+
+        # TF-IDF model
+        from sklearn.feature_extraction.text import TfidfVectorizer
```
arekmano: import at the top

src/nlp_service/outlier/outlier_detection.py
```diff
@@ -0,0 +1,62 @@
...
+        tfidf_vectorizer = TfidfVectorizer(ngram_range=self.NGRAM_RANGE, strip_accents='ascii')
+        X_tfidf = tfidf_vectorizer.fit_transform(all_sentences)
+
+        # Fit to robust covariance estimation
+        from sklearn.covariance import EllipticEnvelope
```
arekmano: import at Top

src/nlp_service/outlier/outlier_detection.py
```diff
@@ -0,0 +1,62 @@
+import glob
+import os
+import json
+import pickle
```
arekmano: use Joblib. more efficient for large arrays.

src/nlp_service/controllers/nlpController.py
```diff
@@ -22,6 +24,10 @@
 rasaClassifier = RasaClassifier()
 rasaClassifier.train(force_train=True)
 
+# Outlier detector - Predicts if the new message is a clear outlier based on a model trained with fact messages
+outlier_detector = OutlierDetection()
```
arekmano, line 28: load from disk using joblib or pickle. Faster than retraining.

arekmano (changes requested):

src/nlp_service/controllers/nlpController.py
lancelafontaine: Instantiating the object doesn't retrain. It's only retrained when calling `initialize_fact_model`, and the latest changes never call that (use a bin instead).

src/nlp_service/outlier/outlier_detection.py
```diff
@@ -0,0 +1,61 @@
...
+        outlier_estimator.fit(X_tfidf.toarray())
+
+        # Binarize for future use
+        with open(self.TFIFD_PICKLE_FILE, 'wb') as f:
+            joblib.dump(tfidf_vectorizer, f, compress=True)
+        with open(self.OUTLIER_PICKLE_FILE, 'wb') as f:
+            joblib.dump(outlier_estimator, f, compress=True)
+
+    def predict_if_outlier(self, sentences):
+        self.ensure_models_loaded()
```
arekmano: wouldn't it make more sense to load the model when the flask app is initialized?

arekmano (changes requested):

src/nlp_service/outlier/outlier_detection.py
arekmano: doing it this was will make the first request wait for the model to be loaded.

arekmano (approved):
LGTM

Issue #46, opened 2017-10-09
Indicate the understood response from the user's answer


arekmano:
*Description*
- As a user, I would like to validate and correct the facts and categories that the system understood

*Scope of Work*
- [x] For 3 sample classifications that the system makes, through NLP, the system must output what it internally understood. (i.e. for NLP, the system would signal the user the classification that it was able to extract from the user input).
- [ ] Create a high-fidelity mock-up of interaction.

*Acceptance Criteria*
- System asks a fact question.
- User enters a textual answer to the system. The response of the system contains a list of labels that it was able to extract from the textual answer, indicating which one the system thinks is the most likely.

*Story Points*
- 8

*Priority*
- High

*Risk*
- Low

naregeff:
Verified, high priority. This will allow training of the AI system, right?

arekmano:
Yes. Also it will allow us to make correct predictions, rather than having incorrect facts about the case. Depending on the UI design, this story and #65 potentially overlap.

arekmano:
Moved to iteration 5, as the UI components have not been worked on.

naregeff:
no high-fi mockup. it's been developed and functional, but UI needs work. can sign off, conditional to adding this to a new "_UI review" story_".

PR #197, opened 2017-11-22
[#46] Performs all modifications necessary on web client for recording user's confirmation to bot responses for training our NLP/ML.

lancelafontaine:
[#46] 

- [x] Added prompt for user confirmation with buttons/inputs once our application has made a prediction based on their input.
- [x] Client  makes HTTP POST request to submit whether our bot's prediction was accurate

These include all modifications necessary on web client for obtaining user data to train our NLP/ML.

### DIFF
--- a/src/backend_service/app.py
+++ b/src/backend_service/app.py
@@ -49,6 +49,10 @@ def handle_files(conversation_id=None):
     else:
         abort(make_response(jsonify(message="Invalid request"), 400))
 
+@app.route("/nlp-response", methods=['POST'])
+def train_nlp_from_user_response():
+    # TODO: Persist the user's confirmation for whether NLP prediction was correct
+    return make_response(jsonify(request.get_json()))
 
 @app.route("/legal", methods=['GET'])
 def get_legal_documents():
--- a/src/web_client/src/components/Chat.vue
+++ b/src/web_client/src/components/Chat.vue
@@ -46,7 +46,20 @@
             <div id="chat-message-zeus">
               <img v-if="!zeus.input" alt="" src="../assets/chatting.gif">
               <transition name="fade">
-                <p v-if="zeus.input" v-html="zeus.input"></p>
+                <div>
+                  <div v-if="zeus.enableUserConfirmation">
+                    <p>Please confirm whether our bot's prediction was correct</p>
+                    <div v-if="factType === 'boolean'">
+                      <button v-on:click="confirmBotResponse(true)">Yes</button>
+                      <button v-on:click="confirmBotResponse(false)">No</button>
+                    </div>
+                    <div v-if="factType !== 'boolean'">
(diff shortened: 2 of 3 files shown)


lancelafontaine:
image 1

dodoels:
1. What if the user answers the question without clicking the button?
2. Is the feedback corresponding to current question?
3. Can you make the feedback independent (like in a bubble somewhere next to the avatar) instead of inside the chat? Since it is more like a feedback system from user perspective, it shouldn't be inside the main conversation flow.

lancelafontaine:
1. The way it is now, we are not forcing the user to enter feedback. They can keep answering questions without giving feedback on bot predictions.
2. Feedback is about the prediction, so it's for the `Oh yes, I know all about problems with lease termination` part. If it's not clear, we can reword it.
3. I didn't spend too much time on this, because this is 100% temporary. It will not be included in the final product. This is just for getting a base of user responses from our beta users. Once that's done, we'll remove (or obfuscate) this feature entirely.

Let me know if that answers your questions.

src/backend_service/app.py
```diff
@@ -49,6 +49,10 @@ def handle_files(conversation_id=None):
     else:
         abort(make_response(jsonify(message="Invalid request"), 400))
 
+@app.route("/nlp-response", methods=['POST'])
+def train_nlp_from_user_response():
+    # TODO: Persist the user's confirmation for whether NLP prediction was correct
```
arekmano, line 54: very important TODO

arekmano:
is the response persistance going to be done in another PR or this one? I noticed the TODO in the code

lancelafontaine:
@arekmano Different PR. These changes only touch the web client. The next PR will concern everything related to backend logic.

lancelafontaine:
However, this PR encompasses all client-side features that would be required to complete this user story (ie. from a strictly client-side point of view, it's feature complete).

src/backend_service/app.py
Vynny: Personally, I think the training should be a scheduled task. It would take very long to retrain rasa at every request imo.

lancelafontaine: @Vynny My intention was not to retrain on every request, but to store the response on every request.

I figured we could then recreate the db based on a postgresl dump, and retrain the model once based on that.

But these changes would be in incorporated in the next PR anyways.

arekmano: i think the function name is perhaps misleading, as it is supposed to store the response, not train the NLP

arekmano (approved):
Approving but persisting response is critical for future

src/backend_service/app.py
lancelafontaine: I agree, naming is hard. Will fix in next PR.

PR #198, opened 2017-11-22
[#183] Added primary ML endpoint to readme.

arekmano:
[#183]

- [x] Added primary ML endpoint to readme


src/ml_service/README.md
```diff
@@ -15,6 +15,127 @@
...
+**Method** : `POST`
+
+**Data constraints**
+
+Provide facts_vector and demands_vector, with key values for each fact/demand.
+
+```json
+{
+  "demands" : {
+    "demand_lease_modification": 1,
```
Samuel-Campbell, line 79: why are all the values 1?

dodoels:
@arekmano it is a bit confusing if you have all `1` as the value, can you change some to `0` instead

src/ml_service/README.md
arekmano: dummy example

arekmano: updated

dodoels (approved):

lancelafontaine:
This is only one of the many endpoints that would be required for integration with `nlp_service` right? `nlp_service` will (eventually) need a way to know what fact to ask the user next.

arekmano:
yes. This is the initial api that we can demo this iteration. we will later need an next_fact endpoint or similar

ghost (changes requested):
Code Cov bug...?

ghost (approved):

PR #205, opened 2017-11-24
[#46] Returns NLP prediction results to front-end if available for user confirmation

lancelafontaine:
[#46] 

- [x] Returns NLP prediction results to front-end if available for user confirmation.

image 2
image 3

### DIFF
--- a/src/backend_service/controllers/conversationController.py
+++ b/src/backend_service/controllers/conversationController.py
@@ -65,6 +65,9 @@ def receive_message(conversation_id, message):
     possible_answers = None
     additional_info = None
     enforce_possible_answer = False
+
+    user_message = None
+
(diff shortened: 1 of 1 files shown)


src/backend_service/controllers/conversationController.py
```diff
@@ -127,6 +130,24 @@ def receive_message(conversation_id, message):
         if enforce_possible_answer:
             response_dict['enforce_possible_answer'] = True
 
+    # Get the last extracted fact entity
+    if len(conversation.fact_entities) > 0 and conversation.fact_entities[-1]:
```
arekmano, line 134: method is exceptionally long. will have to be refactored eventually

arekmano (approved):

Issue #159, opened 2017-11-11
Remove English precedents from the dataset


arekmano:
English precedents are interfering with our ability to cluster properly. As such, we should programmatically remove them from our dataset.

arekmano:
Bug is fixed. 154 english precedents are removed

PR #216, opened 2017-12-07
[#159] Ignoring English precedents

TaimoorRana:
- Ignoring English precedents


ghost (commented):
Why are we ignoring them every time instead of just making a script that detects them, then we manually check to make sure it didn't just match something that one of the parties said in English in quotation marks and THEN we simply remove them?

Issue #217, opened 2017-12-08
Sign up to Procezeus by asking a tenant/landlord legal question


lancelafontaine:
*Description*
As a user, I would like to submit a question about Quebec tenant/landlord law and get a reply in email form.

*Scope of Work*
- [x] Create a mockup for the landing page
- [x] Create a landing page with a form, to enter an email address and legal question
- [x] Add entries to database when user submits them

*Demo requirement*
1. The user navigates to the landing page
2. User types a question and their email
3. A success modal pops up telling them that we will get back to them

*Acceptance Criteria*
- User enters a question and email
- Email + Question are added to the database

**Documentation, files, diagrams, etc. relating to this feature:**
Url, books, etc.

arekmano:
After this user story and #218 are completed, we should focus on marketing this to the public, to collect data about their questions + email.

arekmano:
This story is very similar to #218. The difference being that for this story, users ask a question. In the other, users simply sign up for the beta

arekmano:
Some drawn mockups about potential solutions: ![Overall Mockup](https://i.imgur.com/vMvzDwT.jpg)

Traditional signup:

![Signup form 1](https://i.imgur.com/TH4QCtj.jpg)

Conversational Signup (I personally prefer this):

![Signup form 2](https://i.imgur.com/3iogwpI.jpg)

One more sign up form (obligatory question): 

![Signup form 3](https://i.imgur.com/7K6X6NZ.jpg)

lancelafontaine:
If we're opting for simplicity, I prefer option 3. Option 2 would showcase our product more but we would have to polish up our existing chat interface.

arekmano:
Since the UI functionality of this feature is minimal (restricted conversation, straightforward answers), we can possibly do 2; it would give our app some personality, which could be good with getting actual beta testers in the future.

naregeff:
functionality works.

PR #219, opened 2017-12-18
[#217] Initial Beta Sign up

arekmano:
- Extracted conversation from main chat
- Added Sample conversation flow
- Added state machine for supplying email/question
- TODO: Send question + Email to BE for storage

[Feedback Requested]
[#217]


arekmano:
image

arekmano:
image

dodoels:
So after I talked to @lancelafontaine we came up with an idea how to have all the codes in our repo with minimum duplication. 

1. instead of making a new directory for duplicated front-end infrastructure, we change the config in the existing infrastructure with an environment variable, let's say `CJL_RELEASE`. We build the page within the infra. When the variable is set to `beta`, the CI/CD is gonna pick up the code and build based on the config (i.e. build the beta landing page ONLY, and ignore all other pages and routing)
2. create `beta_backend_service` with independent database
3. our CD is gonna have two configs: 1) it builds everything on the front-end except the beta page, and builds every service except `beta_backend_service`, and associate to `capstone.cjl.com`; 2) it builds only the beta page on FE and only the `beta_backend_service`, and associate to `procezeus.ca`

@arekmano

lancelafontaine:
Clarification for 3: we would still only have on CI/CD pipeline. Builds would run tests for all services (beta, non-beta) and deploy both to our server.

arekmano:
1. That seems like significant effort for a change that we will essentially "dispose" of. 
2. I think it makes sense to have both the beta page up as well as our application. Beta would be for public, application page would be for demos, internal testing, etc.
3. Configuration + pipeline changes do not add user value IMO and we should stick to what works already.

For these reasons I think it makes sense to decouple them "nearly" entirely, and have the beta be a seperate entity. Something like nginx routing should discriminate what we show users. (e.g. procezeus.ca -> beta page, capstone.cjl -> application)

dodoels (approved):
Approved and merge now. PR is too big and can't see the difference. I will work on the UI based on this.

PR #223, opened 2017-12-21
[#159] remove english precedents from raw data

TaimoorRana:
[#159]

- [x] Found and removed 154 English precedents
- [x] manually checked 20 random precedents and they were indeed English


src/ml_service/util/precendent_directory_cleaner.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+    :param directory_path: directory path to search in
+    :param language: language that needs to be removed
+    :return: total files removed
+    """
+    count = 0
+    for filename in os.listdir(directory_path):
+        if filename.endswith(".txt"):
+            file_content = open(directory_path + filename, "r", encoding="ISO-8859-1").read()
+            if is_language_type(file_content, language):
```
arekmano, line 28: if detect(file_content) == language

no need for additional method

TaimoorRana: I agree but It's was more to separate the logic of deleting a file from detecting a certain language.

arekmano: IMO no need to have layers of abstraction. I don't foresee us ever having to reuse this code

dodoels (approved):
LGTM functionally

src/ml_service/util/precedent_directory_cleaner_test.py
```diff
@@ -0,0 +1,10 @@
+import unittest
+import util.precendent_directory_cleaner as directory_cleaner
+from util.constant import Path
+
+
+class TestLogger(unittest.TestCase):
```
Samuel-Campbell: change class name

TaimoorRana: Good catch! thanks

Issue #45, opened 2017-10-09
Accept user feedback within system


arekmano:
*Description*
- As a user, I would like to give feedback on missing features, so that they can later be implemented

*Scope of Work*
- Design Feedback component's UI
- Implement Feedback component
- Write Unit tests demonstrating feedback functionality
 
*Acceptance Criteria*
- User presses a button, types a string and presses submit. The user gets a confirmation popup saying that his feedback is received.
- User feedback strings, their conversation identifiers and question identifier are persisted within the database.

*Story Points*
- 3

*Priority*
- Medium

*Risk*
- Low

arekmano:
clarification is needed: Should the feedback be obtained via an email address (user sends via their Gmail or Outlook client), or should we handle the feedback that is given within our system, i.e. create a feedback form?

naregeff:
Verified.
Priority medium (super important for improving through iterating, but it isn't a core feature).
Feedback handled within the system. UI link, popup form, possibility to include screenshots maybe (encourage / allow user to be as precise as possible).

arekmano:
This story will be needed if we are to have beta testers

arekmano:
Something similar to this should be good:
![](https://i.giphy.com/media/xTiTnlktZPH07oxhYs/giphy.webp)

In short: Give USERS GIFS AND THEY WILL RESPOND!

lancelafontaine:
This is for users to send us feedback in the form of free form text.

Samuel-Campbell:
User feedback strings, their conversation identifiers and question identifier are persisted within the database.

I don't understand this. Why not just leave it anonymous

naregeff:
works! we'll make it look sexier and add gifs.

PR #224, opened 2017-12-21
[#45] User Feedback Feature

Samuel-Campbell:
[#45]

- [x] Adds ability for users to submit feedback


Samuel-Campbell:
screenshot from 2017-12-21 09-17-32
screenshot from 2017-12-21 09-17-46
screenshot from 2017-12-21 09-17-53

arekmano:
UI LGTM

src/postgresql_db/models.py
```diff
@@ -154,6 +154,13 @@ class File(db.Model):
     timestamp = db.Column(db.DateTime, default=datetime.datetime.utcnow)
 
 
+class Feedback(db.Model):
+    id = db.Column(db.Integer, primary_key=True)
+
+    # Attributes
+    feedback = db.Column(db.String(250), nullable=False)
```
Vynny: This should be db.Text() if we aren't imposing length limits.

src/web_client/src/components/Sidebar.vue
```diff
@@ -82,13 +97,32 @@ export default {
...
+            dialogVisible: false,
             username: this.$localStorage.get('username').toUpperCase(),
             //TODO: fetch username from conversation, now use usertype instead
             api_url: process.env.API_URL,
             connectionError: false
         }
     },
     methods: {
+        submitFeedback(){
+            var feedback = document.getElementById("feedback-text").value;
```
Vynny: This does not leverage vue.js. You should do this by either passing the value into the function on submit, or add a feedback object to 'data' then bind it to the input with v-model. Then in the method you can do 'this.feedback.text' to retrieve the text.

src/web_client/src/components/Sidebar.vue
```diff
@@ -82,13 +97,32 @@ export default {
...
+            this.$http.post(this.api_url + 'feedback',{
+                data: feedback
+            }).then(
+                response => {
+                    alert("Thank you for providing us with feedback!");
+                    this.dialogVisible = false;
+                    document.getElementById("feedback-text").value = "";
```
Vynny: Same as above, this should be bound to a model and not done with plain js.

Vynny (changes requested):

src/web_client/src/components/Sidebar.vue
Vynny: You can see an example of this here: 
https://stackoverflow.com/a/45840995
https://vuejs.org/v2/guide/forms.html

src/web_client/src/components/Sidebar.vue
arekmano: +1

src/web_client/src/components/Sidebar.vue
```diff
@@ -82,13 +97,32 @@ export default {
             uploadedFileList: new Array,
             openFileList: false,
             openReportList: false,
+            dialogVisible: false,
```
dodoels: `openFeedbackModal` for consistent naming convention (start with verb)

src/web_client/src/components/Sidebar.vue
```diff
@@ -61,6 +61,21 @@
                 <h3>PREVIOUS CASES</h3>
             </div>
             <!-- End of Previous Case List -->
+            <!-- Feedback -->
+            <div id="sidebar-feedback" class="sidebar-menu" @click="dialogVisible = true">
+                <h3>FEEDBACK</h3>
+            </div>
+            <!-- el-dialog for feedback -->
+            <el-dialog title="Feedback" :visible.sync="dialogVisible">
+                <textarea id="feedback-text">
```
dodoels: 1. in `data ()`, in `user` object, create new value `feedback` 
2. add `v-model="user.feedback"`

src/web_client/src/components/Sidebar.vue
```diff
@@ -82,13 +97,32 @@ export default {
...
             //TODO: fetch username from conversation, now use usertype instead
             api_url: process.env.API_URL,
             connectionError: false
         }
     },
     methods: {
+        submitFeedback(){
+            var feedback = document.getElementById("feedback-text").value;
+            this.$http.post(this.api_url + 'feedback',{
+                data: feedback
```
dodoels: 3. `data: this.user.feedback`

src/web_client/src/components/Sidebar.vue
dodoels: 4. `this.user.feedback == null`

Vynny (approved):

dodoels (approved):
LGTM, I will add unit test and improve the code a bit later

Issue #232, opened 2017-12-22
French Sign Up page


arekmano:
*Description*
As a user, I would like the system to show me a french beta sign up page after i change the language

*Scope of Work*
- [x] Translate strings to french
- [x] Create option to toggle between french and english

*Demo requirement*
1. The user toggles to french
2. The system shows the french version of the beta sign up

*Acceptance Criteria*
- The user is able to x
- The user sees y
- The user hears z

arekmano:
Current Transcript:

> **zeus**:'Hello there stranger! My name is Zeus and I'm here to assist you with your Québec tenant/landlord issues. I'll be going into beta soon, and would like to get more context about common problems. If you don't mind telling me about any tenant/landlord issues you have, we'll try to get back to you as soon as possible with some useful information.'
> **user**: 'Okay'
> **zeus**:'Would you like to ask me a question about tenant/landlord issue?'
> **user**: 'Sure'  OR 'I have no question'
> **zeus**:'Great! What is your question?'
> **user**: 'Confirm' + QUESTION
> **zeus**:'Mind leaving your email so that we can let you know once our beta is live?'
> **user**: 'Confirm' + EMAIL
> **zeus**:'Would you like to receive updates on our beta test?'
> **user**: 'Sure' OR 'No, thanks'
> **zeus**:'Thank you so much! I'll be in touch. Have a nice day!'
> **user**: 'Close'

arekmano:
Proposed changes to transcript:

> **zeus**: **'Hello there stranger!'**
> **zeus**: **'My name is Zeus and I'm here to assist you with your Canadian tenant/landlord issues.'**
> **zeus**: **'If you don't mind telling me about any tenant/landlord issues you have, I'll try to get back to you as soon as possible with some useful information.'**
> **user**: 'Okay'
> **zeus**: **'Do you have a legal question for me?'**
> **user**: 'Sure'  OR 'I have no question'
> **zeus**:'Great! What is your question?'
> **user**: 'Confirm' + QUESTION
> **zeus**:'Mind leaving your email so that we can let you know once our beta is live?'
> **user**: 'Confirm' + EMAIL
> **zeus**:'Would you like to receive updates on our beta test?'
> **user**: 'Sure' OR 'No, thanks'
> **zeus**:'Thank you so much! I'll be in touch. Have a nice day!'
> **user**: 'Close'

arekmano:
Transcript validation needed

dodoels:
[...]

PR #239, opened 2017-12-27
[#232] Added french input buttons and text fields to beta page

arekmano:
[#232]

- [x] Added french input buttons and text fields to beta page

### DIFF
--- a/src/beta_client/src/components/Chat.vue
+++ b/src/beta_client/src/components/Chat.vue
@@ -28,7 +28,7 @@
             <el-input v-if="answer.type=='question' && language == 'en'" autosize v-model="userQuestion" placeholder="ENTER YOUR QUESTION" autoComplete="off"></el-input>
             <el-input v-if="answer.type=='email' && language == 'en'" autosize v-model="userEmail" placeholder="ENTER YOUR EMAIL"></el-input>
             <el-input v-if="answer.type=='question' && language == 'fr'" autosize v-model="userQuestion" placeholder="DEMANDEZ VOTRE QUESTION" autoComplete="off"></el-input>
-            <el-input v-if="answer.type=='email' && language == 'fr'" autosize v-model="userEmail" placeholder="ENTRER VOTRE COURRIEL"></el-input>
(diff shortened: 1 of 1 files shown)


arekmano:
French input fields and buttons:
image

*spelling mistake was fixed

arekmano:
Pinging for review @Vynny @lancelafontaine @Taimoorrana1 @mihaiqc @choitwao

src/beta_client/src/components/Chat.vue
```diff
@@ -25,11 +25,14 @@
       <el-row>
         <div v-for="answer in currentConversation.user">
           <el-col v-if="answer.type=='question' || answer.type=='email'" :sm="24">
-            <el-input v-if="answer.type=='question'" autosize v-model="userQuestion" placeholder="ENTER YOUR QUESTION" autoComplete="off"></el-input>
-            <el-input v-if="answer.type=='email'" autosize v-model="userEmail" placeholder="ENTER YOUR EMAIL"></el-input>
+            <el-input v-if="answer.type=='question' && language == 'en'" autosize v-model="userQuestion" placeholder="ENTER YOUR QUESTION" autoComplete="off"></el-input>
+            <el-input v-if="answer.type=='email' && language == 'en'" autosize v-model="userEmail" placeholder="ENTER YOUR EMAIL"></el-input>
```
lancelafontaine, line 29: I think it would be better for us to have an object like so:

```json
{
    "EN": {
        "answerText": "english"
    },
    "FR": {
        "answerText": "francais"
    },
}
```

That way, changing the language is just assigning either the `EN` or `FR` object to a variable, and having the two-way binding in Vue.js auto-update everything (instead of having conditionals that check for language in lots of `divs`.).

Let's keep this in mind when we apply internationalization/localization to our main application.

lancelafontaine (approved):

src/beta_client/src/components/Chat.vue
arekmano: How would that work? We would still need a conditional language check to see what placeholder text we display?

arekmano: @lancelafontaine the "language" variable is already performing this task

Issue #162, opened 2017-11-11
Remove previously understood fact


arekmano:
*Description*
As a user, I would like to erase previously understood facts from the conversation's list

This story has a prerequisite of story #161

*Scope of Work*
- [ ] Create high fidelity UI mockup
- [ ] Add new question generation logic to NLP

*Acceptance Criteria*
- User answers fact 1, fact 2, fact 3 questions
- User then presses on the remove button associated with the fact 1 question's answer
- Fact 1 is removed from the conversation
- User gets an updated question, based on the current set of facts.

lancelafontaine:
We should clarify whether or not the ML will need to recompute what facts to ask about next once a fact is deleted by the user. If that's the case, we need to make a corresponding user story for this functionality.

mihaiqc:
ML's job is to produce facts and spit them out to NLP. All it has to do, logic has to be assured by them imo. So I don't really understand what the backend has to do in this. All backend should do is communicate to the frontend what to show, in this specific scenario.

arekmano:
That is not a priority within this story IMO. We should keep it as simple as possible (i.e. ask the question they just removed)

Vynny:
Are we just re-asking the question they removed, or is there a way to tell the ml (now or at some point) that a particular fact has a null value and shouldn't be taken into consideration?

arekmano:
ATM we should just re-ask the same question. We can refine the implementation later on should the need arise.

naregeff:
Functionality demonstrated.

PR #243, opened 2018-01-04
[#162] Adds fact removal endpoint

Vynny:
[#162]

- [X] Adds backend endpoint to remove a fact that has been resolved previously during the conversation
- [X] Adds the fact entity id to the json data from the various endpoints that return fact entities 
- [X] Fact will be attempted to be resolved again during the conversation once deleted
- [X] Changes API docs to reflect new content returned from endpoints


src/nlp_service/controllers/nlpController.py
```diff
@@ -161,6 +156,7 @@ def classify_fact_value(conversation_id, message):
 
 def __classify_claim_category(message):
     classify_dict = rasaClassifier.classify_problem_category(message)
+    print(classify_dict)
```
arekmano: remove print statement

arekmano (approved):
Print statement in code. Remove. Other than that, looks good.

Issue #237, opened 2017-12-23
DEV STORY: Research and implement production monitoring service


arekmano:
*Description*
As a developer, I would like be notified when there are issues with our production systems.

*Scope of Work*
- [ ] Define extent and sophistication of  monitoring and health checks
    - Come up with list of scenarios as "failure states" which should trigger a response from our monitoring solution
- [ ] Research adequate solutions which can address all failure states
- [ ] Implement monitoring solution
- [ ] Ensure that a failure state is able to notify us (through GitHub or Slack)

*Demo requirement*
1. Monitoring system has been able to report a failure state to the team.

*Acceptance Criteria*
- The team is informed when a failure state in production has occurred.

lancelafontaine:
An account was creation at https://uptimerobot.com/ with the procezeus@gmail.com email.

lancelafontaine:
We've integrated the alerting as a channel notification in our Slack group

![image](https://user-images.githubusercontent.com/1548285/34653787-43820fde-f3bf-11e7-8981-fb1d681d37f3.png)

naregeff:
awesome, now we know when things break.

PR #244, opened 2018-01-05
[#237] Fixes errors in starting up project

lancelafontaine:
[#237]

- [x] Explicitly specifies where the DB volume resides on the host's filesystem
- [x] Fixes some indent changes that prevented boot up

### DIFF
--- a/docker-compose.base.yml
+++ b/docker-compose.base.yml
@@ -48,7 +48,7 @@ services:
     image: postgres
     restart: always
     volumes:
-      - ./data:/var/lib/postgresql/data
+      - ~/.procezeus/data:/var/lib/postgresql/data
     networks:
       - procezeus_net
 networks:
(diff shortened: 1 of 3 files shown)


Vynny (approved):
LGTM. As long as the Arch nemesis is conquered.

docker-compose.base.yml
```diff
@@ -48,12 +48,9 @@ services:
     image: postgres
     restart: always
     volumes:
-      - procezeus_db:/var/lib/postgresql/data
+      - ~/.procezeus/data:/var/lib/postgresql/data
```
arekmano, line 51: Why don't we let Docker manage the container volumes?

lancelafontaine: Latest docker (at least on my machine) seems to throw errors when attempting to create a directory for the local volume. This workaround resolves that issue.

PR #245, opened 2018-01-07
[#237] Health endpoints

lancelafontaine:
[#237]

- [x] Creates a simple health endpoint (200 OK, no body) for the beta and main backend services

### DIFF
--- a/src/backend_service/app.py
+++ b/src/backend_service/app.py
@@ -18,6 +18,11 @@ CORS(app)
 from controllers import conversationController, legalController, feedback_controller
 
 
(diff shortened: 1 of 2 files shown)


arekmano:
Why not simply ping the main page and existing endpoints?

lancelafontaine:
Plan was to ping both homepage as well as backend service (to ensure that nginx is serving files as well as acting as a proxy to a running service).

These new endpoints add very little additional load and don't cause any side effects.

arekmano (approved):
lgtm

Issue #235, opened 2017-12-23
Adding FAQ Questions from RDL


arekmano:
*Description*
As a user, I would like the system to respond to questions listed in the Regie du Logement FAQ

*Scope of Work*
- [x] Scrape all question/answers from https://www.rdl.gouv.qc.ca/en/faqs
- [x] Group common scenarios, and split into different "fact" values. i.e. the following questions:
"If a tenant refuses the rent increase asked by the landlord, can he force the tenant to move?" and "Can a tenant refuse the rent increase asked by the landlord?" can map to facts "refuse_to_move" and "refuse_rent_increase"
- [x] Possibly limit to this user story to the top 10 FAQ questions

*Demo requirement*
1. User asks a question found in the FAQ
2. The system answers with the correct answer

*Acceptance Criteria*
-

naregeff:
works!

PR #249, opened 2018-01-10
[#235] Respond to FAQs with static strings

lancelafontaine:
[#235] 

- [x] Adds very basic ability to recognize FAQ intents and respond with a static string


src/nlp_service/rasa/data/category/claim_category.json
```diff
@@ -2438,6 +2438,51 @@
             "entity": "terminate"
           }
         ]
+      },
+      {
+        "text": "I'm not able to attend the hearing.",
```
ghost, line 2479: We should not just jam the FAQ inside of claim category. We should make a separate JSON file and a separate model IMO or else it will get ultra hard to keep track of.

lancelafontaine: So I agree with that fact that this file is way too long and unstructured. However, we are still effectively determining a claim category, so however we structure it, it should still be logically structured with other claim categories imo.

mihaiqc: I strongly suggest we use a different JSON file for a separate model or we're going to have to keep track of every individual FAQ intent every time we add one to the model since we want a different behavior for them (AKA repeat "What's your problem?" question if the user wants to ask the machine something else).

FAQ Questions are meant for quick info, there is no ML involved AKA different behavior than problem category.

Having a separate model would make adding FAQ questions way more modular in future endeavors and easier to keep track of.

ghost: Shouldn't we use a different JSON file for a separate model or we're going to have to keep track of every individual FAQ intent every time we add one to the model since we want a different behavior for them (AKA repeat "What's your problem?" question if the user wants to ask the machine something else and not show claim_validity, etc.)

Having a separate model would make adding FAQ questions way more modular in future endeavors and easier to keep track of.

lancelafontaine: Sure. But ultimately, when the user types their question to us, we'll need to run it through _both_ models, to determine whether they are asking a FAQ question or a question we'll need to do ML on.

We can separate the models from each other for organizational purposes, but we can't avoid running through both models on that first question, since we have no idea what the user may want.

lancelafontaine: Having two models may also complicate our % diff technique we are using now, but that's something we can try to deal with later.

ghost: Correct, we run it through both models. The lowest intent confidence gets filtered out. If the chosen one is the FAQ one, it will have nothing to do with the ML.

I just honestly feel that adding everything into one JSON is a band-aid solution. Especially once we put in more FAQs from Educaloi or Likehome.info
Then we add in entity extraction.

ghost: I honestly don't think we'll be able to keep up with "which is which" as we add-on features if we don't make this separation.

lancelafontaine:
image

ghost (approved):

lancelafontaine:
image

src/nlp_service/rasa/data/category/claim_category.json
lancelafontaine: Added the `FAQ_RLQ` identifier to the intents so that we're able to originate the response for those intents.

arekmano: Since both the FAQ question and claim category are determined by the same question, it makes sense to have them in the same model

arekmano (approved):
lgtm

PR #266, opened 2018-01-13
[#254] Fixed chat scroll function so that latest bubble is always sho…

arekmano:
…wn at the same location

[#254]


dodoels:
Did you test responsive of your changes? The element selector is too barebone to me but for beta I can live with it :)

Issue #261, opened 2018-01-12
Show precedents that are similar to the current claim


arekmano:
*Description*
As a user, I would like the system to show me the 5 most similar precedents to my claim

*Scope of Work*
- [x] Investigate if there are similarity metrics that are easily calculable given our SVM weights (Probably not, but still worth looking). [Update: nearest neighbour calculation required additional distance calculation, as the SVM's weights are only related to answering a single specific question] 
- [x] Implement 5-NN algorithm using sci-kit learn
- [x] Create UI response that outputs the case number of the precedents that are most similar, as well as a similarity metric (% similar or distance measure)

*Acceptance Criteria*
- The user fills out claim facts
- Chatbot responds with top 5 most similar case numbers

arekmano:
This user story likely involves finding the 3 or 4 nearest neighbors of the data point, under a threshold distance value

dodoels:
Are we allow to show the precedents?

Vynny:
As @choitwao mentioned, are we allowed to show this data?
I'm uncertain that showing users case numbers is going to offer them any value, unless we somehow summarize every case and present that on the front end.

Perhaps just offering a count? 
Example: "I know of 4 cases that are similar to yours"

arekmano:
just display the case numbers as a POC. This we are 100% fine with displaying.

arekmano:
Precedent similarity was implemented using euclidean distance, on the assumption that smaller distance means that the data points are more closely related.

This assumption holds with features that are binary. However, as <a href="https://link.springer.com/chapter/10.1007/3-540-44503-X_27">this paper</a> indicates, this assumption may not always hold true. We will have to reevaluate the approach once our features are converted to values as part of #251.

arekmano:
One possible solution to the problem could be to use Mahalanobi distance, rather than euclidean, which would take into account the variance of each feature, therefore reducing the effect of large values (i.e. monthly dollar costs)

Relevant Paper:
http://www.sciencedirect.com/science/article/pii/S0169743999000477

arekmano:
In order to evaluate the best distance metric, manual precedent evaluation must be performed.

After this, we can determine how accurate the measurement is in detecting similarity

arekmano:
[...]

PR #277, opened 2018-01-18
[#261] Implemented precedent similarity using mahalanobi distance

arekmano:
- Returns the 5 most similar cases by the case's features (demands, facts, outcomes)
- The better our feature extraction is, the better the similarity metric will be.

error. commit message says that it uses KNN, yet this is false. Algorithm computes distance metric for all points and returns the ones with smallest distance. (see http://scikit-learn.org/stable/modules/generated/sklearn.neighbors.NearestNeighbors.html#sklearn.neighbors.NearestNeighbors.kneighbors)

[#261]


Samuel-Campbell (approved):

Samuel-Campbell (commented):
Good idea to pre train the similar cases

arekmano:
In order to evaluate the best distance metric, manual precedent evaluation must be performed.

After this, we can determine how accurate the measurement is in detecting similarity

Issue #251, opened 2018-01-12
Claim: Predict $ that may be obtained


arekmano:
*Description*
As a user, I would like the system to predict how much money a landlord may obtain from their tenant in a standard case.

The predicted amount should consider:
- Rent payment
- Compensatory damages

*Scope of Work*
- [x] Regex for $ amounts pertaining to Rent payment
- [x] Refactor post_processing module to accommodate integers 
- [x] Refactor Regex dictionary
- [x] regex for outcomes 
- [x] Update readme with command lines
- [ ] Unittest regexlib

*Demo requirement*
1. The user does x
2. The system does y
n. Rest of a use case of a flow chart

*Acceptance Criteria*
- The user sends a list of facts concerning their case
- The system responds with a monetary amount that they may be entitled to

Samuel-Campbell:
Progress so far
--------------------------------------------
INFO: Total precedents parsed: 44970
INFO: Total precedents with absent                                    : 1779
INFO: Total precedents with apartment_impropre                        : 563
INFO: Total precedents with apartment_infestation                     : 0
INFO: Total precedents with asker_is_landlord                         : 31567
INFO: Total precedents with asker_is_tenant                           : 3081
INFO: Total precedents with bothers_others                            : 3
INFO: Total precedents with case_fee_reimbursement                    : 1226
INFO: Total precedents with disrespect_previous_judgement             : 138
INFO: Total precedents with incorrect_facts                           : 26
INFO: Total precedents with landlord_inspector_fees                   : 513
INFO: Total precedents with landlord_notifies_tenant_retake_apartment : 384
INFO: Total precedents with landlord_pays_indemnity                   : 41
INFO: Total precedents with landlord_prejudice_justified              : 18574
INFO: Total precedents with landlord_relocation_indemnity_fees        : 673
INFO: Total precedents with landlord_rent_change                      : 2385
INFO: Total precedents with landlord_rent_change_doc_renseignements   : 311
INFO: Total precedents with landlord_rent_change_piece_justification  : 310
INFO: Total precedents with landlord_rent_change_receipts             : 310
INFO: Total precedents with landlord_retakes_apartment                : 1784
INFO: Total precedents with landlord_retakes_apartment_indemnity      : 95
INFO: Total precedents with landlord_sends_demand_regie_logement      : 582
INFO: Total precedents with landlord_serious_prejudice                : 129
INFO: Total precedents with lease                                     : 26289
[...]

PR #278, opened 2018-01-18
[#251] updated readme, updated main.py's comments, updated console ou…

Samuel-Campbell:
[#251]

1) Updated readme
2) updated console output for debugging
3) fixed certain comments
4) try/except uncaught error when using pre-processor
5) added sys.exit() statements if certain files are missing


arekmano (approved):
LGTM

Issue #268, opened 2018-01-14
Tweak data to give better more consistent results for problem categorization


ghost:
*Description*
As a user, I would like the system to give better results for my problem categorization sequence.

*Scope of Work*
- [x] Create quick tool to assess confidence per experimental sentence (quick .py file)
- [x] Add/Modify common examples and Regex depending on varying results
- [x] Try to reproduce and fix the "number crash" ("31258182" crashed NLP apparently, I was NOT able to reproduce this bug)

*Demo requirement*
1. The user does enters their problem on the website
2. The system adequately identifies the category from the input string
3. System doesn't crash when "312412541" is an input inside of the text label

*Acceptance Criteria*
1. The user does enters their problem on the website
2. The system adequately identifies the category from the input string
3. The user attempts to input a number such as "13251235236" where a yes/no value or a problem category is expected and the system doesn't crash.

**Documentation, files, diagrams, etc. relating to this feature:**
Url, books, etc.

***Changes***:
- [Optional]

arekmano:
moving to iteration 8

Vynny:
We may need to do some form of user input normalization to get better results.

For instance, currently the sentence "I'm being evicted" will properly classify ask "lease_termination" with about 70% confidence. However, "Im being evicted" does not classify with a confidence of only 44%.

Essentially, we need to find some way to handle contractions and synonyms to get better classification results.

Vynny:
This task is likely going to be larger than expected as it cannot be solved by simply tweaking the rasa training data. I recommend that this be reclassified from bug to story.

lancelafontaine:
We should also consider adding integration tests to this story to ensure that all provided sample sentences end up being classified to their proper intent. We shouldn't consider it acceptable for example sentences to fail to be classified.

ghost:
Multiple solutions were found to tweak the RASA data to give back results that abide by our expectations.
1. Converting our data set to use lowercases increased the intent confidence percentage considerably
2. Adding common examples and Regexes increased the intent confidence percentage but was mostly work on issue #280 as our intent confidence ratio dropped back considerably while working on LikeHome.info FAQ
3. The bug reported by our team was never reproduced afterwards despite our best efforts and could therefore not be solved.

naregeff:
Accuracy got one step closer.

PR #281, opened 2018-01-18
[#268] Allows cjl script to run tests per service instead of all tests at once

lancelafontaine:
[#268] 

- [x] Allows cjl script to run tests per service instead of all tests at once


Vynny (approved):
YES

PR #283, opened 2018-01-18
 [#268] Runs all common example sentences against intents for NLP integration tests

lancelafontaine:
[#268]

- [x] Runs all common example sentences, ensures that percent difference works, and queries them against their intents as an integration test


Vynny (approved):
It's good but we should add some ad-hoc examples too in the future.

arekmano (approved):
much needed

PR #296, opened 2018-01-22
[#251] multilabel classifier

Samuel-Campbell:
[#251]

1) Added multi class classifier
2) Added unittest
3) Revised every single ML unittest and fixed some of them
4) ready to train


src/ml_service/model_training/abstract_models/abstract_classifier.py
```diff
@@ -0,0 +1,43 @@
+import numpy as np
+from util.log import Log
+from sklearn.metrics import precision_recall_fscore_support
+
+
+class AbstractClassifier:
```
arekmano: IMO this is unnecessary Abstraction. While the classifiers may have the same methods, their implementations can vary. I.e. get_weights is assuming that Scikit will be used. Same with test. These effectively block off Neural networks from being used (Tensorflow, Keras, etc)

arekmano (dismissed):

src/ml_service/model_training/abstract_models/abstract_classifier.py
Samuel-Campbell: You can just overwrite methods. I essentially wanted us to use a template

dodoels (approved):
Look clean to me, can't find any issue.

src/ml_service/model_training/abstract_models/abstract_classifier.py
arekmano: Still. Inheritance does not make much sense in the Python language. I will dismiss my review, but still stand by my original argument

Issue #260, opened 2018-01-12
Revamp Chatbot UI for primary application


arekmano:
*Description*
As a user, I would like the primary chatbot to resemble the sign up UI

*Scope of Work*
- [x] convert the chat method to #264 along with chat history shown in the interface
- [x] remove chat history overlay and leave only recognized fact (table); convert such overlay to right toggle sidebar
- [x] technical debt: prompt user's name
- [x] toggle left sidebar
- [x] Multiple sentences in zeus' response

dodoels:
I will be working on this issue along with #162 
Before I updated the description, here is what I understand for revamping the UI:
- change the color scheme based on PO's input
- convert the chat method to #264 along with chat history shown in the interface
- remove chat history overlay and leave only recognized fact (table); convert such overlay to right toggle sidebar
- technical debt: prompt user's name
- toggle left sidebar 
- responsive design

arekmano:
meeting may be required to discuss these, but yes.

naregeff:
main app UI consistent with revamped signup UI.

PR #298, opened 2018-01-23
[#260] UI 

dodoels:
[#260]

- finalize chatbot interface

having issue on scrolling, will fix it later. 
Need to update data population when api is updated (for multiple sentences).

image
image


Samuel-Campbell (approved):

arekmano:
what about multiple statements in a row?

PR #299, opened 2018-01-23
[#260] Removes debugging output for most recently categorized fact value

lancelafontaine:
[#260]

- [x] Removes debugging output for most recently categorized fact value
- [x] `response_text` is always parsed as HTML so no need for multiple key values


lancelafontaine:
screenshot from 2018-01-23 09-53-22

arekmano (approved):
lgtm

Vynny (approved):
LGTM

PR #300, opened 2018-01-23
[#251] SVR Regression

Samuel-Campbell:
[#251]

1) Finished regressoin
2) made unittests
3) tested on sample size of 100

### DIFF
--- a/src/ml_service/model_training/regression/multi_class_svr.py
+++ b/src/ml_service/model_training/regression/multi_class_svr.py
@@ -1,0 +1,118 @@
+import numpy as npr
+from sklearn.svm import SVR
+from sklearn.model_selection import train_test_split
+from util.file import Load, Save
+from util.log import Log
+from feature_extraction.post_processing.regex.regex_tagger import TagPrecedents
+
+
+class MultiClassSVM:
+
+    def __init__(self, data_set=None):
+        """
+        Constructor
+        :param data_set: [{
+            name: 'AZ-XXXXXXX.txt',
+            demands_vector: [...],
+            facts_vector: [...],
+            outcomes_vector: [...]
+        },
+        {
+            ...
+        }]
+        """
+        self.data_set = data_set
+        self.model = None
+
+    def get_weights(self):
+        """
+        The weight associated with each input fact.
+        Useful in seeing which facts the classifier
+        values more than others
+        :return: None
+        """
+        if self.model is not None:
+            return self.model.coef_[0]
+        Log.write('Please train or load the classifier first')
+        return None
+
+    def __test(self, x_test, y_test):
+        """
+        UPDATE
+        :param x_test:
+        :param y_test:
+        :return: None
+        """
+        pass
+
+    def train(self):
+        """
+        CHANGE
+        :return: None
+        """
+        x_total, y_total = self.reshape_dataset() # 1
+        self.mlb = MultiLabelBinarizer() # 2
+        y_total = self.mlb.fit_transform(y_total)
+
+        x_train, x_test, y_train, y_test = train_test_split(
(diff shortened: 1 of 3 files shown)


src/ml_service/model_training/regression/multi_class_svr.py
```diff
@@ -0,0 +1,143 @@
...
+    def __test(self, x_test, y_test):
+        """
+
+        :param x_test:
+        :param y_test:
+        :return: None
+        """
+        Log.write("Testing Classifier")
+        y_predict = self.model.predict(x_test)
+        r_Score = r2_score(y_test, y_predict)
```
arekmano: it makes more sense to have an R2 score and explained variance per regressor, (i.e. per outcome predicted) rather than the overall score, as we can identify which outcomes are having difficulties with a particular regressor.

src/ml_service/model_training/regression/multi_class_svr.py
```diff
@@ -0,0 +1,143 @@
...
+            name: 'AZ-XXXXXXX.txt',
+            demands_vector: [...],
+            facts_vector: [...],
+            outcomes_vector: [...]
+        },
+        {
+            ...
+        }]
+        """
+        self.data_set = data_set
```
arekmano, line 31: since the dataset is huge, doing this would make the MultiClassSVR instance unnecessarily large, either train in the constructor, or clear the dataset field after the training is complete.

src/ml_service/model_training/regression/multi_class_svr.py
```diff
@@ -0,0 +1,143 @@
...
+        return self.model.predict([data])
+
+    def load(self):
+        """
+        Returns a model of the classifier
+        :return: OneVsRestClassifier(SVC())
+        """
+        self.model = Load.load_binary("multi_class_svr_model.bin")
+
+    def reshape_dataset(self):
```
arekmano, line 145: make this a private method? (non-blocker)

src/ml_service/model_training/regression/multi_class_svr.py
```diff
@@ -0,0 +1,143 @@
...
+
+        x_train, x_test, y_train, y_test = train_test_split(
+            x_total, y_total, test_size=0.20, random_state=42) # 2
+
+        Log.write("Sample size: {}".format(len(x_total)))
+        Log.write("Train size: {}".format(len(x_train)))
+        Log.write("Test size: {}".format(len(x_test)))
+        Log.write("Training Regression Using Multi Class SVR")
+
+        clf = MultiOutputRegressor(SVR(kernel='linear')) # 3
```
arekmano, line 117: There should add a method similar to SVM#evaluate_best_parameters, to brute force the best possible kernel, C, and epsilon parameters of the SVR. This would help us isolate the best parameters for our dataset. Additionally, using MultiOutputRegressor will cause all predictions to be given by the same estimator, which will not give the most optimal outcome.

I.e. additional_indemnity_outcome and tenant_ordered_to_pay_landlord may have better results with "linear" and "poly" kernels respectively

arekmano (changes requested):

src/ml_service/model_training/regression/multi_class_svr.py
arekmano: this is a non-blocker but a code change will have to be made if we want to present good results to the PO

Samuel-Campbell: No read this:
scikit-learn.org/stable/modules/generated/sklearn.multioutput.MultiOutputRegressor.html

MultiOutputRegressor generates a different formula for each column

arekmano: explained further in slack

arekmano (approved):
lgtm

PR #301, opened 2018-01-23
[#251] Linear Regression

dodoels:
[#251]

- implement linear regression
- I manage to run the regression on all our data and see if the metrics is good enough compare against SVR
- unit test will be later


src/ml_service/model_training/regression/linear.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def save(self):
+        save = Save()
+        save.save_binary("linear_regression_model.bin", self.model)
+
+    def predict(self, data):
+        return self.model.predict([data])
+
+    def load(self):
+        self.model = Load.load_binary("multi_class_svm_model.bin")
```
arekmano, line 46: save and load are different binaries

src/ml_service/model_training/regression/linear.py
```diff
@@ -0,0 +1,61 @@
...
+class LinearRegression:
+
+    def __init__(self, data_set=None):
+        self.data_set = data_set
+        self.model = None
+
+    def __test(self, x_test, y_test):
+        Log.write("Testing Linear Regression")
+        y_predict = self.model.predict(x_test)
+        r_Score = r2_score(y_test, y_predict)
```
arekmano, line 20: Do not combine the outcome predictors for the r2 score. a much more meaningful number is the r2 and explained variance per outcome prediction

arekmano (changes requested):

arekmano:
What are the metrics associated with Linear Regression usage? sample size, r2, variance, etc.

src/ml_service/model_training/regression/linear.py
dodoels: Yeah not ready for merge yet, need to rename/documentation/test

dodoels: also linear regression might have bad performance, need to test it out with more data. This PR is mainly for discussion purpose. I will do some research on polynomial regression and make a comparison

arekmano: I don't see the difference between this method and the method used by @Samuel-Campbell. His SVR implementation uses a linear kernel, which means that this will have the same properties as your approach.

https://stackoverflow.com/questions/46970888/difference-between-linearregression-and-svm-svrkernel-linear

One tool that is worth looking into is GridSearchCV (found here: http://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html). This would allow us to test multiple different regression methods and determine the best one. See LinearSVM#evaluate_best_parameters within our application for an example

PR #303, opened 2018-01-24
[#251] refactor ml_controller and added unittest

Samuel-Campbell:
[#251]

1) refactored ml controller
2) works with new models
3) unittest


src/ml_service/web/ml_controller.py
```diff
@@ -1,69 +1,105 @@
 from feature_extraction.post_processing.regex.regex_tagger import TagPrecedents
-from flask import abort
 from model_training.classifier.multi_class_svm import MultiClassSVM
+from model_training.regression.multi_class_svr import MultiClassSVR
 import numpy as np
-from util.constant import Path
 
-linear_classifier = MultiClassSVM()
-linear_classifier.load()
-indices = TagPrecedents().get_intent_indice()
```
dodoels, line 6: Maybe rename `/web/` to just `/controller/`, and `MLController` to `PredictController`

src/ml_service/web/ml_controller.py
```diff
@@ -1,69 +1,105 @@
...
+        integer_outcome_vector = MlController.regression_model.predict(int_facts_vector)
 
+        bool_facts_vector = MlController.dict_to_bool_vector(input_json['facts'])
+        binary_outcome_vector = MlController.classifier_model.predict(bool_facts_vector)
 
-def __dict_to_vector(input_dict, input_type):
-    """
+        return MlController.vector_to_dict(binary_outcome_vector, integer_outcome_vector)
+
+    @staticmethod
```
dodoels, line 45: why do we need bool_vector and int_vector for different models? Both regression and classification should take the same vector and predict two distinct sets of outcomes no?

Samuel-Campbell: we have 2 prediction models. 
1 for regression problems and 1 for classification problems
Both models will return a different matrix. I need to then join them

src/ml_service/web/ml_controller.py
Samuel-Campbell: Let's keep the names as is because they are standardized in the code

src/ml_service/web/ml_controller.py
```diff
@@ -1,69 +1,106 @@
...
 
+        bool_facts_vector = MlController.dict_to_bool_vector(input_json['facts'])
+        binary_outcome_vector = MlController.classifier_model.predict(bool_facts_vector)[0]
 
-def __dict_to_vector(input_dict, input_type):
-    """
+        return MlController.vector_to_dict(binary_outcome_vector, integer_outcome_vector)
+
+    @staticmethod
+    def dict_to_bool_vector(input_dict):
```
arekmano: This is odd. I would not expect the classifier to output anything other than a 1 or 0. So why can the value be >= 1?

Samuel-Campbell: the inputs are 1 and 0. I can explain in person. I'll be at school soon

Samuel-Campbell: ohh because i receive [1, 0 , 5000, ...] from the NLP service. I binarize it.  but i use sklearn.preprocessing for that now. Just didn't push yet

src/ml_service/web/ml_controller.py
dodoels: I get what you mean, but you are not explaining why you need two different vector for different models. Our models will deal with the boolean and int issues before training, for me this is an extra step we dont need.

Samuel-Campbell: i refactored the code. look again

TaimoorRana:
Looks good

TaimoorRana (approved):

PR #321, opened 2018-01-29
[#251] regex lib test --- NOT READY TO MERGE

TaimoorRana:
1. Added fact and demand cluster to  CJL server
2. Download and unzip the cluster folder and add them to their appropriate location
3. Run regexes against each cluster file to find a matching cluster file
4. Save all the matches in a dict and binarizes it 
5. Load this dict in the test file to check if the regex does indeed match the corresponding cluster file


src/ml_service/init.py
```diff
@@ -21,6 +21,10 @@
     'non-lem': ['https://capstone.cyberjustice.ca/data/frWac_non_lem_no_postag_no_phrase_200_skip_cut100.bin', 126052447],
     'multi_class_svm_model': ['https://capstone.cyberjustice.ca/data/bin/multi_class_svm_model.bin', 11434717],
     'classifier_labels': ['https://capstone.cyberjustice.ca/data/bin/classifier_labels.bin', 275],
+    'multi_class_svr_model': ['https://capstone.cyberjustice.ca/data/bin/multi_class_svr_model.bin', 22977311],
```
Samuel-Campbell: remove. not used anymore

src/ml_service/init.py
```diff
@@ -21,6 +21,10 @@
...
+    'svr_scaler_model': ['https://capstone.cyberjustice.ca/data/bin/svr_scaler_model.bin', 2623],
```
Samuel-Campbell: remove. not used anymore

src/ml_service/init.py
```diff
@@ -21,6 +21,10 @@
...
+    'fact_cluster': ['https://capstone.cyberjustice.ca/data/bin/fact_cluster.zip', 13337193],
```
Samuel-Campbell: remove. unnecessary download

src/ml_service/init.py
```diff
@@ -21,6 +21,10 @@
...
+    'decision_cluster': ['https://capstone.cyberjustice.ca/data/bin/decision_cluster.zip', 1209757],
```
Samuel-Campbell: remove. unnecessary download

TaimoorRana: fact and decision cluster are needed. removing the others

Issue #186, opened 2017-11-15
Claim: Retaking the rental


arekmano:
*Description*
As a landlord, I would like the system to predict whether or not retaking my rental is justified

*Scope of Work*
- [x] Fix/update regexes
- [x] Create supervised learning algorithm benchmark for evaluating predictions
- [x] Implement supervised learning algorithm that can evaluate whether or not retaking my rental is justified based on facts present in the case .
- [x] Extract using regex, or similar technique whether or not  the rental was retaken in a specific case.
- [x] Integrate with NLP

*Demo requirement*
1. The user selects the facts that are relevant to their case
2. The user sends the facts to the system
3. The system displays "Retaking the rental is justified" or "Retaking the rental is not justified" 

*Acceptance Criteria*
- The user is able perform demo steps
- The F1 score of the supervised learning algorithm is >50%.

arekmano:
Moving this out of iteration, as higher priority user stories have come in

naregeff:
Demo not possible. Need to figure out questions, and integrate to NLP.

Vynny:
Was not integrated with NLP. Need to go over fact specifics with Sam prior to integration.

naregeff:
Works, would need to add some legal rules for more accurate predictions.

PR #326, opened 2018-02-02
[#186] multi output regression

Samuel-Campbell:
1) Added multi output regressor wrapper
2) will be easier to make many predictions at once
3) NEED TO UPLOAD NEW MODELS ONCE MERGED ** NOT TO SELF

Every column  has a label. From this label we know if we expect an int or boolean value. Iterate each column and if the classifier determined the outcome was present then calculate its integer value.


arekmano (commented):
This change will conflict with #327

src/ml_service/model_training/regression/multi_output_regression.py
```diff
@@ -0,0 +1,63 @@
...
+
+        2) If 1) is true, then replace the value in the vector by it's true integer
+           value.
+
+        Additional notes: step 2) only applies when the boolean value of the classifier
+                          returns True. The reason is that the regressor is trained on biased
+                          inputs. In a sense, the classifier serves a filtering layer to know
+                          which fields are needed by the regressor.
+
+        :param data: np.array([1, 0, 1, 1, ...])
```
TaimoorRana: need to update

src/ml_service/model_training/regression/multi_output_regression.py
```diff
@@ -0,0 +1,63 @@
...
+        :return: np.array([1, 0, 22, 2, ...])
+        """
+        for i in range(len(outcomes)):
+            if outcomes[i] == 1:
+                column_name = MultiOutputRegression.classifier_labels[i][0]
+
+                if column_name == 'additional_indemnity_date':
+                    pass
```
Taimoorrana1: why is this needed?

arekmano (approved):
Looks good. I don't think we have a case there the data is not biased on win/loss so this approach would definitely work for all the project predictions

dodoels (approved):

PR #327, opened 2018-02-03
[#261] Added API integration, removed demands from similarity

arekmano:
- Added ML service API integration for similiarity finder
- Removed demands from similarity finder, as we are never supplied with it
- Changed the distance metric from mahalanobi to euclidean with a scaler, as mahalanobi requires a covariance matrix, which we cannot calculate with our dataset (memory requirements are too high)

### DIFF
--- a/src/ml_service/model_training/similar_finder/similar_finder.py
+++ b/src/ml_service/model_training/similar_finder/similar_finder.py
@@ -1,5 +1,6 @@
 from util.file import Save, Load
 from sklearn.neighbors import NearestNeighbors
+from sklearn.preprocessing import StandardScaler
 import numpy as np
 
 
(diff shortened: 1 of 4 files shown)


src/ml_service/model_training/similar_finder/similar_finder.py
```diff
@@ -17,19 +18,23 @@ def __init__(self, train=False, dataset=[]):
...
+            sample_set = [np.concatenate(
+                [vec['facts_vector'], vec['outcomes_vector']]) for vec in dataset]
 
             for i in range(len(sample_set)):
                 sample_set[i] = sample_set[i].astype(np.int64)
 
-            self.model = NearestNeighbors(5, metric='mahalanobis', metric_params={'V': np.cov(sample_set)})
+            self.scaler = StandardScaler()
+            sample_set = self.scaler.fit_transform(sample_set)
+            self.model = NearestNeighbors(5, metric='euclidean')
```
Samuel-Campbell, line 31: scaler + euclidean distance makes sense

src/ml_service/web/ml_controller.py
```diff
@@ -31,8 +33,10 @@ def predict_outcome(input_json):
         """
         facts_vector = MlController.dict_to_vector(input_json['facts'])
         outcome_vector = GlobalPredictor.predict_outcome(facts_vector)
-
-        return MlController.vector_to_dict(outcome_vector)
+        response = MlController.vector_to_dict(outcome_vector)
+        similar_dict = {'facts_vector': facts_vector, 'outcomes_vector': outcome_vector}
```
Samuel-Campbell, line 37: I think we should make a different POST request for this. We are not predicting an outcome here

arekmano: I included it in the /predict endpoint since the NLP service will only be calling this at the end, to obtain the overall results page. SimilarFinder uses the same inputs and will be shown at the same time as the predictions, so it makes sense to include them in the same endpoint.

Samuel-Campbell: ohhh right got'cha. It's true that the inputs are the same

Samuel-Campbell (approved):

src/ml_service/model_training/similar_finder/similar_finder.py
```diff
@@ -40,7 +45,7 @@ def __init__(self, train=False, dataset=[]):
             e.g. {
```
TaimoorRana, line 45: Need to move the comments underneath the method header as we are doing this for all the other files. Not a blocker

TaimoorRana (approved):
Looks good. Just have a look at my comment

Issue #25, opened 2017-09-28
Isolate the Document from a picture and align it


dodoels:
*Description*
- As I user, I want to be able to use my lease to obtain facts about my claim

*Scope of Work*

*Acceptance Criteria*

arekmano:
Does this story still apply? Would it be relevant to turn this story into "As I user, I want to be able to use my lease to obtain facts about my claim"? What pieces of information are we extracting from the lease/bill of sale? What is a minimum set of attributes that we can extract?

naregeff:
Yeah, I added two distinct user stories on the Google Sheet and tagged them with "consumer" and "logement".

We can look at this together: I'll bring a lease with me.

Edit: keep in mind, lease form is mandatory.

arekmano:
As part of this story, document segmentation and image transforming was completed.

What remains:
- Actual OCR
- Document Segmentation

naregeff:
Signing off on this part, it works!

lancelafontaine:
![lease_tilted](https://user-images.githubusercontent.com/1548285/36054583-4590bb98-0dc5-11e8-95d9-220ef985100e.png)
![lease_straight](https://user-images.githubusercontent.com/1548285/36054584-45c34dc4-0dc5-11e8-887f-81667f178a2b.png)

PR #329, opened 2018-02-04
[#25] Implements simplest-case OCR in task_service

lancelafontaine:
[#25] 

- [x] Installs and configures Tesseract + OpenCV on Arch Linux Docker image
- [x] Implements simplest case and tests for OCR of an aligned black-and-white image with no distortions


src/task_service/app.py
```diff
@@ -22,5 +22,8 @@
 @app.route("/ocr/extract_text", methods=['POST'])
 def extract_text():
     request_data = request.get_json()
-    return ocr_controller.extract_text(request_data['conversation_id'], request_data['image_data'])
+    if request_data and 'conversation_id' in request_data and 'image_data' in request_data:
+        return ocr_controller.extract_text(request_data['conversation_id'], request_data['image_data'])
+    return abort(make_response(jsonify(message="JSON must contain 'conversation_id' and 'image_data' keys."), 422))
```
dodoels, line 27: image as JSON?

src/task_service/controllers/ocr_controller.py
```diff
@@ -16,9 +18,19 @@ def extract_text(conversation_id, image_data):
...
+    img = _get_image_from_data(image_data)
+    text = _get_string_from_np_img(img)
     return jsonify({
-        'image_text': image_data
+        'image_text': text
     })
 
-if __name__ == '__main__':
-    extract_text(1, 'this is image data')
+def _get_image_from_file(file_path):
```
dodoels, line 27: Is this in used?

src/task_service/app.py
dodoels: nvm, i don't think you have api ready

dodoels (approved):

src/task_service/controllers/ocr_controller.py
lancelafontaine: Currently no, but it might make sense to use in the future (?)

src/task_service/app.py
lancelafontaine: The api "should" work, although it's not  hooked up with anything else.

`image_data` is just the image data as a string. This is how the backend file upload endpoint worked too.

Issue #315, opened 2018-01-27
NLP Data Revamp


ghost:
*Description*
As a user, I would like the system to categorize my problem more effectively.

*Scope of Work*
- [x] Cut RLD RLQ in favor of LikeHome when intersecting
- [x] Cut fluff sentences
- [x] Separate FAQ questions and problem categories such that they are relevant only to either landlord or tenant person types

*Acceptance Criteria*
- The user is able to pinpoint their claim category with more accuracy (less errors)
- A landlord is only able to get answers to questions that are relevant to them (FAQ & Claim Category)
- A tenant is only able to get answers to questions that are relevant to them (FAQ & Claim Category)

ghost:
@naregeff Priority level? (basically increasing the intent confidence with which it predicts how to answer- this is way necessary before adding more NLP datapoints)

ghost:
NLP is now revamped to take into account context-sensitive answers depending on the nature (landlord or tenant) of the user using the application.

A tenant will get answers relevant to him while a landlord will get an answer relevant to his situation.

ghost:
Notable ~100-130% increase in intent confidence.

naregeff:
I'm very proud of you, this works great!

PR #332, opened 2018-02-05
[#315] Enabled comments in custom json scripts

Samuel-Campbell:
Any line starting with '#' will be commented

example: #comment
example: # comment
example: \s\s\s\s\s\s\s # \s\s\s\\s\s\s\s\s\       comment


Vynny (commented):
Could you make it so that's it's only considered a comment if it starts with a '#', rather than being anywhere in the line?

Vynny (approved):
LGTM

arekmano (approved):

dodoels (approved):

PR #334, opened 2018-02-06
[#315] Splitting claim categories between landlord and tenant person types

Vynny:
[#315]

- [X] Separates landlord and tenant claim categories
- [X] Adds context sensitive responses to FAQ questions, based on whether user is landlord or tenant
- [X] Reworks training data for claim classification to be more like something either a landlord or tenant would write
- [X] Updates regex features to give better classification results
- [X] Removes rent change claim category and some faq questions that were either irrelevant or better handled by existing ML functionality
- [X] Re-enables NLP integration test, which now tests both landlord and tenant claim category classification


src/postgresql_db/models.py
```diff
@@ -191,27 +190,33 @@ def get_or_create(session, model, **kwargs):
     {'name': 'asker_is_tenant', 'summary': 'Tenant', 'type': FactType.BOOLEAN},
```
Samuel-Campbell, line 190: is it normal that none of the facts are int?

Samuel-Campbell (approved):
looks fine

src/postgresql_db/models.py
Vynny: yeah, we don't have an int fact type
there's MONEY though, see tenant_owes_rent

PR #336, opened 2018-02-06
[#261] Integrates precedent similarity into prediction statement

Vynny:
[#261] 

- [X] When a prediction is made and similar precedents are returned by the ml service, they will be displayed to the user.
- [X] Similar precedents are displayed in ascending distance order.
- [X] Changes the format in which similar precedents are returned by ml service to make them more readable.

similar_precedents

### DIFF
--- a/src/nlp_service/services/response_strings.py
+++ b/src/nlp_service/services/response_strings.py
@@ -46,6 +46,12 @@ class Responses:
         "missing_category": ["Sorry, I cannot yet make a prediction for this claim category yet."]
     }
 
+    similar_outcome = [
+        "By the way, I know of {} cases similar to yours.|They are {}.",
+        "I've found {} cases that are a lot like yours.|I know them as {}.",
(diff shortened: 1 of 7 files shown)


src/nlp_service/services/response_strings.py
```diff
@@ -484,27 +490,39 @@ def faq_statement(claim_category_value, person_type):
...
         for prediction in prediction_dict:
             prediction_value = prediction_dict[prediction]
             prediction_predicate = int(prediction_value) > 0
             prediction_text = Responses.chooseFrom(
                 Responses.prediction[claim_category_value][prediction][prediction_predicate]).format(prediction_value)
-            all_predictions.append(prediction_text)
+            all_responses.append(prediction_text)
+
+        # Generate similar outcome text
+        if len(similar_precedent_list) > 0:
```
arekmano, line 516: Currently, it returns 5 most similar precedents, it does not threshold the distance. We can implement on the ML side if this causes problems though

arekmano (approved):
LGTM :+1:

PR #338, opened 2018-02-07
[#315] Yes/No fix for NLP Data optimization

ghost:
- [x] Fixing yes/no classification with data sets add-on
- [x] Updating RASA NLU from 11.0 to 11.2
- [x] Updating Duckling from 1.7.2 to 1.7.3


Vynny (approved):
LGTM

arekmano (approved):
LGTM

dodoels:
So the issue was caused by no enough data?
Also what is the change in RASA update and why we updated it?

dodoels (approved):

PR #339, opened 2018-02-07
[#0] Adds a *destructive* command in our cjl script to reset the database

lancelafontaine:
[#0]

- [x] Adds a *destructive* command in our `cjl` script to reset the database on the local host


Vynny (approved):
LGTM

cjl
```diff
@@ -21,6 +21,18 @@ if [ "$1" == "clean" ]; then
   exit
 fi
 
+if [ "$1" == "reset-db" ]; then
+  echo "WARNING: This command is destructive. It removes all database data on your local machine."
+  read -r -p "Are you SURE you want to continue? [y/N]" clean_confirmation
```
arekmano, line 26: wouldn't this make it harder to automate?

arekmano: i.e. for deployments?

dodoels: probably `echo Y`?

lancelafontaine: Would make sense to have a `-y` or `--noconfirm` flag to skip this check entirely.

PR #341, opened 2018-02-07
[#315] Fixes true, false misclassifications for several facts

Vynny:
[#315]

- [X] Fixes data sets of facts not generated from the base yes-no training data which would sometimes miss-classify yes no answers
- [X] Adds gender regex to some of the training data

In a future iteration, this should be addressed by adding an import feature to the json creator tool.


ghost (approved):

arekmano (approved):
Lgtm

Issue #346, opened 2018-02-09
I want to answer more questions to obtain a more accurate prediction


arekmano:
*Description*
As a user I want the system to only ask me the most important questions in order to resolve my case.

*Scope of Work*
- [x] Seperate important from non important questions 
- [x] Establish anti-facts
- [x] Integrate with NLP

*Demo requirement*
1. The user answers a set of important questions for a claim category
2. The user is given a prediction
3. The user is asked whether they would like to continue answering questions to get a better prediction
4. If yes, the user can answer 5 more questions at a time. After 5 questions a new prediction will be given.
5. If more additional facts remain, goto step 3, else end the conversation.

*Acceptance Criteria*
- The user is required to answer a set of important questions
- The user is prompted to answer additional questions once given a prediction

naregeff:
It works! 👍 we proposed an alternate workflow where the user can stop the flow of questions anytime after 5 and request a prediction.

PR #348, opened 2018-02-15
[#346] seperated important from non important facts

Samuel-Campbell:
1) Updated README
2) Facts which have the most weight are seperated from those which don't
2.1) Calculate mean of all weights
2.2) Find it's order of magnitude of the mean
2.3) all facts with greater or equal order of magnitude are important
2.4) all facts with lower order of magnitude are not important
2.5) facts with weight of 0 are ignored
3) Anti-facts are also mapped

### DIFF
--- a/src/ml_service/README.md
+++ b/src/ml_service/README.md
@@ -204,16 +204,38 @@ None
             "tenant_is_bothered",
             "bothers_others"
         ]
-    },
-    "anti_facts": {
-        "tenant_rent_not_paid_less_3_weeks": "tenant_rent_not_paid_more_3_weeks",
-        "tenant_lease_fixed": "tenant_lease_indeterminate",
-        "not_violent": "violent",
-        "tenant_individual_responsability": "tenant_group_responsability"
-    }
+    }   
 }
 ```
 
+### Get Anti Facts
+
+Get the anti facts
+
+Left hand side always initialized to 1 and right hand side to 0
+
+**URL**: `/antifacts`
+
+**Method**: `GET`
+
+**Data constraints**
+
+None
+
+#### Success Response
+
+**Code** : `200 OK`
+
+**Content examples**
+
+```json
+{
+    "tenant_rent_not_paid_less_3_weeks": "tenant_rent_not_paid_more_3_weeks",
+    "tenant_lease_fixed": "tenant_lease_indeterminate",
+    "not_violent": "violent",
+    "tenant_individual_responsability": "tenant_group_responsability"
+}
+```
 ---
 
 
--- a/src/ml_service/README.md
+++ b/src/ml_service/README.md
@@ -162,7 +162,7 @@ Provide facts_vector and demands_vector, with key values for each fact/demand.
 
 **Code** : `400 Bad Request` - *Inputs not provided*
 
-**Code** : `404 Not Found` - *Conversation doesn't exist*
+**Code** : `404 Not Found` - *Conversation doesn"t exist*
 
 ### Get Fact Weights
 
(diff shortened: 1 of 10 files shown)


src/ml_service/README.md
```diff
@@ -184,20 +184,33 @@ None
...
+            "tenant_damaged_rental",
+            "tenant_individual_responsability",
+            "signed_proof_of_rent_debt",
+            "tenant_lease_indeterminate",
+            "tenant_dead",
+            "tenant_is_bothered",
+            "bothers_others"
+        ]
+    },
+    "anti_facts": {
```
Vynny: Would it be possible to make a separate endpoint for the anti facts? Because you're mixing the outcome keys like "additional_indemnity_money" with meta data "anti_facts". 
Either that or you can nest them such that it looks more like this:

```
{
	"outcomes": {
		"additional_indemnity_money": {
			"important_facts": [],
			"not_important_facts": []
		}
	},
	"anti-facts": {
		"tenant_rent_not_paid_less_3_weeks": "tenant_rent_not_paid_more_3_weeks",
		"tenant_lease_fixed": "tenant_lease_indeterminate",
		"not_violent": "violent",
		"tenant_individual_responsability": "tenant_group_responsability"
	}
}
```

Vynny (changes requested):

src/ml_service/README.md
Samuel-Campbell: yes you're right. That was the original plan and I forgot. I'll just make 2 end points

src/ml_service/model_training/classifier/multi_output/multi_class_svm.py
```diff
@@ -90,11 +104,17 @@ def get_ordered_weights(self):
...
             weights = estimator.coef_[0]
             for j in range(len(weights)):
-                outcome_list.append([labels['facts_vector'][j][1], weights[j]])
+                if weights[j] > 0:
+                    outcome_list.append([labels['facts_vector'][j][1], weights[j]])
 
             outcome_list.sort(key=lambda x: abs(x[1]), reverse=True)
-            outcome_list = [x[0] for x in outcome_list]
-            weight_dict[self.classifier_labels[i][0]] = outcome_list
+            weights = [abs(x[1]) for x in outcome_list]
```
arekmano, line 111: This is not a good way of ranking our weights. 

If weight A is -9x10^-1  and weight B is 3x10^0, then weight B is more important than A.

A more accurate approach would be to take e^weight

src/ml_service/model_training/classifier/multi_output/multi_class_svm.py
```diff
@@ -90,11 +104,17 @@ def get_ordered_weights(self):
...
+            mean_power = math.log10(np.mean(np.array(weights)))
+            important_facts = [x[0] for x in outcome_list if math.log10(abs(x[1])) >= mean_power]
```
arekmano, line 113: We should still have some arbitrary minimum number of facts that we ask the user. (i.e. 5). Not sure if that minimum number would fit here or in the NLP service.

arekmano (dismissed):

src/ml_service/model_training/classifier/multi_output/multi_class_svm.py
arekmano: (not blocker) but team input required @Cyberjusticelab/justiceaiteam

Vynny: NLP would rely entirely on what's given to us from the endpoint, so there should be a minimum number of facts per outcome.

src/ml_service/model_training/classifier/multi_output/multi_class_svm.py
Samuel-Campbell: I don't understand what you're saying. B is in fact more important and this is what my algorithm does. log10(weight) and e^weight are inversely proportional

src/ml_service/model_training/classifier/multi_output/multi_class_svm.py
Samuel-Campbell: You don't need this. It's mathematical --> There will always be weights with magnitude above the mean's magnitude. Worst case is that ALL weights are important. There will never be 0 facts

Vynny (approved):

src/ml_service/model_training/classifier/multi_output/multi_class_svm.py
arekmano: I know, but we also don't want 1 question to be asked.

arekmano: Better have a minimum number of questions.
its entirely possible that the weights are
100
5
4
3
2

where 100 would be the only one above mean

src/ml_service/model_training/classifier/multi_output/multi_class_svm.py
arekmano: since these are log probabilities, they can be negative. Taking the absolute value will skew the ordering of the weights. Also, I'm not sure why log10 of the mean is being taken.

Samuel-Campbell: no it won't. -100 has more weight than 10.
Only the absolute value is necessary

Samuel-Campbell: the mean is taken as a threshold

src/ml_service/model_training/classifier/multi_output/multi_class_svm.py
Samuel-Campbell: We can add it later if we see this occur. Furthermore the user can always ask for more questions.. So i dont see the problem.

Vynny:
@arekmano merging, fact weights can be improved later in the iteration. At the moment this is issue is a big blocker.

src/ml_service/model_training/classifier/multi_output/multi_class_svm.py
Samuel-Campbell: Doing e^(weight) will actually yield the opposite of what we want. It is inversely proportional

arekmano: For some reason, I assumed coefficients were on a log scale.

PR #352, opened 2018-02-19
[#346] Integrates ML weight & anti-fact endpoint with NLP, changes NLP question flow

Vynny:
[#346]

- [X] On ML side, renames "not_important_facts" to "additional_facts" for the weight endpoint return dict key
- [X] Updates facts used on NLP after fact refactoring
- [X] NLP now uses ML weight endpoint to retrieve which facts to resolve for a claim category
- [X] NLP now uses ML anti-fact endpoint to automatically resolve antifacts before sending for a prediction
- [X] Adds a rasa acknowledgement classifier. Classifies phrases such as "yeah sure", "not interested" etc in order to determine whether a user wants to answer more questions for a better prediction
- [X] NLP flow changed such that user will be asked all 'important' facts before being given a prediction
- [X] NLP flow changed such that after important facts are asked, user will have the option to answer additional facts for a better prediction. Five additional facts (can be changed by setting MAX_ADDITIONAL_FACT in nlp_controller) will be asked before a new prediction is given and the user is re prompted for whether they want to continue.
- [X] Added a loopback to claim category determination. If a user does not wish to answer more questions, the bot will ask them for another problem. 

**Note: Requires DB reset to function**

flow1
flow2
flow3


dodoels (approved):

src/nlp_service/controllers/nlp_controller.py
```diff
@@ -106,45 +111,128 @@ def classify_fact_value(conversation_id, message):
...
-        # Retrieve the Fact from DB
-        if new_fact_id:
-            new_fact = db.session.query(Fact).get(new_fact_id)
-
-            # Set current_fact to new_fact (returned from ML service)
-            conversation.current_fact = new_fact
-
-            # Generate question for next fact (returned from ML service)
-            question = Responses.fact_question(new_fact.name)
+    ############################
```
TaimoorRana: This function is pretty big, try breaking it down if possible

src/nlp_service/controllers/nlp_controller.py
```diff
@@ -106,45 +111,128 @@ def classify_fact_value(conversation_id, message):
...
+
+        if fact_entity_value is not None:
+            next_fact = fact_service.submit_resolved_fact(conversation, current_fact, fact_entity_value)
+            new_fact_id = next_fact['fact_id']
+
+            if new_fact_id:
+                new_fact = db.session.query(Fact).get(new_fact_id)
+                conversation.current_fact = new_fact
+
+                if fact_service.has_important_facts(conversation):
```
TaimoorRana: I don't see why 2 'if' are needed.

TaimoorRana: `if fact_service.has_important_facts(conversation):` followed by `if new_fact:`

Samuel-Campbell (approved):

PR #353, opened 2018-02-19
[#346] updated regex metrics

Samuel-Campbell:
1) Fixed lines tagged % --> Was previously not counting the matches per statements but per file (mistake)
2) All untagged statements are saved to text file. Those can be reviewed for better regex


dodoels (approved):
LGTM

Issue #44, opened 2017-10-09
Show Conversation Progress %


arekmano:
*Description*
- As a user, I would like to see the progress of the conversation until an outcome is displayed

*Scope of Work*
- [x] Create high-fidelity mock-up of progress module
- [x] Implement progress visual component
- [x] Calculate number of interactions required (either average or specific to the current interaction) on the back-end
 
*Acceptance Criteria*
- When user starts an interaction with the system, the progress is at 0%.
- As the user interacts with the system, the progress continually increases.
- At the result display screen, the progress bar shows as 100%

naregeff:
Scope of work hasn't been defined yet, but seems relatively simple. 
Low priority (but should be done).

arekmano:
Odd, i remember writing the description for this one. GitHub bug.

dodoels:
@Vynny can you return the remainning number of question or percentage to conversation end point?

Vynny:
Yes but full implementation will have to wait for #346 to be complete since it is revamping how facts are determined for a claim. I could however add a dummy return to the API so that you can start working on it.

Samuel-Campbell:
#346 is almost complete @choitwao

Vynny:
I'm taking ideas on how to implement this.
We have important facts and non important facts. 
All important facts must be asked before the first prediction is given.
Once the first prediction is given, the user is asked whether they want to answer more questions. If they accept, 5 (max) non important facts are asked before another prediction is given.
If another prediction is given and non important facts remain, the user will be prompted another time as to whether or not they want to continue.

A) The progress can be set to 100% once the important facts are asked. If they wish to answer more questions, the progress will be set such that once they answer the next batch of questions it will be set to 100% again. 

B) Simple questions-answered/total-questions ratio.

Vynny:
Went with option A, more details can be seen at PR #354.

naregeff:
Works like a charm 👍

PR #354, opened 2018-02-20
[#44] Adds conversation progress to messaging endpoint

Vynny:
[#44]

- [X] Adds progress percentage to backend POST /conversation endpoint

Documented in backend_service README.md. Here is a short summary:
1. Progress is only returned when chatting. It is not an attribute of the conversation. Thus will not be present when fetching the conversation directly by id.
2. The progress is null if there is no fact resolution required. Therefore when a claim category is being determined or when the user has asked a FAQ question, progress will be null. This can be used to determine whether or not to show a progress bar.
3. Progress will be a whole number (%), with 100% triggering a new prediction. 
4. If additional facts are asked, the % is reduced, and a new prediction will be given when it reaches 100%.

### DIFF
--- a/src/nlp_service/controllers/nlp_controller.py
+++ b/src/nlp_service/controllers/nlp_controller.py
@@ -93,7 +93,8 @@ def classify_claim_category(conversation_id, message):
         response = Responses.chooseFrom(Responses.clarify).format(previous_question="")
 
     return jsonify({
-        "message": response
+        "message": response,
+        "conversation_progress": __calculate_conversation_progress(conversation)
     })
 
 
@@ -131,7 +132,8 @@ def classify_fact_value(conversation_id, message):
     db.session.commit()
 
     return jsonify({
-        "message": question
+        "message": question,
+        "conversation_progress": __calculate_conversation_progress(conversation)
     })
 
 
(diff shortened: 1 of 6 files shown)


src/nlp_service/controllers/nlp_controller.py
```diff
@@ -289,6 +291,38 @@ def __state_giving_prediction(conversation):
...
+    Then decreased if additional facts should be answered.
+    """
+    conversation_progress = 0
+    important_fact_count = len(fact_service.get_category_fact_list(conversation.claim_category.value)["facts"])
+    resolved_important_fact_count = fact_service.count_important_facts_resolved(conversation)
+
+    if conversation.bot_state is BotState.GIVING_PREDICTION or conversation.bot_state is BotState.AWAITING_ACKNOWLEDGEMENT:
+        conversation_progress = 1
+    elif conversation.bot_state is BotState.DETERMINE_CLAIM_CATEGORY:
+        conversation_progress = None
```
dodoels, line 308: Can you just set it to 0 as well?

Vynny: I mentioned why I did this in point number 2. It'll be null when you don't have to show a progress bar. No need to show one when we aren't asking a series of questions.

Vynny: For example, lets say they ask a question that can be answered by the FAQ. There's no need to show a progress bar because it's answered immediately and the bot just waits for a new question to be asked. The progress as null serves as an indication to the front-end that progress is irrelevant. Whereas a progress of 0% shows that there is progress because there is going to be a subsequent question relating to their problem.

Samuel-Campbell (approved):

src/nlp_service/controllers/nlp_controller.py
dodoels: the `null` is good for me, but not sure about the FAQ without progress bar. FAQ or claim are distinct from our perspective, not from users'

Vynny: I was under the impression that the progress bar was only for when we were asking a number of questions to give a prediction. There's no point in indicating progress when you get an answer immediately.

Issue #350, opened 2018-02-15
Additional Indemnity Money


Samuel-Campbell:
**Description**
As a user, I would like to know how much additional indemnity money I can receive from a particular case.

**Scope of Work**
- [ ] Add unit tests for regexes
- [ ] Add anti-facts
- [ ] Create supervised learning algorithm benchmark for evaluating predictions
- [ ] Implement supervised learning algorithm that can evaluate whether or not the user is entitled to additional indemnity money
- [ ] Extract using regex, or similar technique whether or not the individual received additional indemnity money in a specific case.

**Demo requirement**

- The user selects the facts that are relevant to their case
- The user sends the facts to the system
- The system displays legal fees in integer form. "You are eligible to 560$."

**Acceptance Criteria**
- The user is able perform demo steps
- The F1 score of the supervised learning algorithm is >50%.

Documentation, files, diagrams, etc. relating to this feature:
Url, books, etc.

Changes:
[Optional]

naregeff:
It seems that the functionality is well implemented, but will require some review by a legal expert to fine tune predictions. 👍

PR #356, opened 2018-02-22
[#350] Abstract Regression

Samuel-Campbell:
1) made this badboy to speed up development of other deep learning regressions
2) Implemented abstract class of regression
3) input dimensions now variable
4) Only need to code 2 methods per ML model


src/ml_service/model_training/regression/single_output_regression/abstract_regressor.py
```diff
@@ -0,0 +1,100 @@
...
+from sklearn.pipeline import Pipeline
+import os
+import numpy as np
+from util.log import Log
+
+
+class AbstractRegressor:
+
+    def __init__(self, regressor_name, dataset = None, outcome_index = 0):
+        """
```
TaimoorRana, line 15: Need to update the comments

dodoels (approved):
Syntax wise LTGTM

src/ml_service/model_training/regression/single_output_regression/abstract_regressor.py
```diff
@@ -0,0 +1,100 @@
...
+    def save(self):
+        """
+            Saves the scaler and regressor. Does not use joblib
+            for the regressor as it is not supported
+        """
+        regressor_name = self.regressor_name
+        file_path = os.path.join(Path.binary_directory, '{}_regressor.bin'.format(regressor_name))
+        Log.write("saving" + '{}_regressor.bin'.format(regressor_name) + " to: " + file_path)
+        Log.write('{}_regressor.bin'.format(regressor_name) + " saved to: " + file_path)
+        self.model.steps[1][1].model.save(file_path)
```
arekmano, line 61: This call is specific to KerasRegressors. it will not work if another library is used, i.e. Sci-kit.

src/ml_service/model_training/regression/single_output_regression/abstract_regressor.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def load(self):
+        """
+            Loads the regressors different components
+        """
+        regressor_name = self.regressor_name
+        Log.write("Loading " + '{}_regressor.bin'.format(regressor_name))
+        file_path = os.path.join(Path.binary_directory, '{}_regressor.bin'.format(regressor_name))
+        Log.write('{}_regressor.bin'.format(regressor_name) + " is successfully loaded")
+        regressor = load_model(file_path)
```
arekmano, line 72: Keras Specific. Are we assuming all our Regressors will be Neural Networks?

arekmano (dismissed):
I don't see the value of abstracting the regressor. We can just have same method names with same args among different gressor implementations.

Issue #345, opened 2018-02-09
DEV STORY: Fix Continuous Deployment


arekmano:
*Description*
As a developer, I would like for my changes to be deployed to our prod-like environment after merging to the master branch.

This feature was previously implemented but unexpectedly broke.

*Scope of Work*
- [x] Investigate reason for CD unexpectedly breaking
- [x] Fix CD to work on every merge to master
- [x] Address any remaining issues.

*Acceptance Criteria*
- Production environment is updated.
- Production environment remains functional after update.

naregeff:
It works again 👍

naregeff:
It works again 👍

PR #359, opened 2018-02-23
[#345] Removes Docker images and database on every deploy

lancelafontaine:
[#345]

- [x] Modify `cjl` script to accept `-y` flags to prevent input prompt on the `clean` and `reset-db` commands.
- [x] Applies both modified `cjl` commands to continous deployment strategy


lancelafontaine:
Since each service is running in Travis as it's own job in parallel, we may need some kind of homemade semaphore that make sure that this only happens once, for the last job to be completed. May require a future PR once this one is merged.

dodoels (approved):

Vynny (approved):

cjl
```diff
@@ -7,27 +7,46 @@ fi
...
     echo "  >>> Killing all docker containers <<<" && echo "" && docker ps -qa | xargs docker kill && echo ""
     echo "  >>> Removing all stopped containers <<<" && echo "" && docker ps -qa | xargs docker rm -v && echo ""
     echo "  >>> Removing all images <<<" && echo "" && docker images -qa | xargs docker rmi && echo ""
+  }
 
+  if [ "$2" != "-y" ]; then
+    echo "WARNING: This command is destructive. It removes all containers and images installed on your local machine."
+    read -r -p "Are you SURE you want to continue? [y/N]" clean_confirmation
+    if [[ $clean_confirmation =~ ^(y|Y)$ ]]; then
+      clean
```
arekmano, line 20: why call it 2x?

arekmano (changes requested):

cjl
```diff
@@ -7,27 +7,46 @@ fi
...
+      clean
+    else
+      echo "Clean aborted."
+    fi
   else
-    echo "Clean aborted."
+      clean
```
arekmano, line 26: same as above

cjl
lancelafontaine: Probably not needed, but I've sometimes some situations where more images get removed if you do it a second time, I think in cases when there are "nested images" (?). Regardless, no harm in running it twice, and our script should try to enforce the state we want.

arekmano (approved):
ok

PR #360, opened 2018-02-24
[#350] improved prediction by averaging fact values

Samuel-Campbell:
1) created a binary dictionary of average vector per regressor
2) Regressor initiates all 0 values to their mean column value before making a prediction
3) fact values greater than 0 are kept as such
4) prediction makes more sense given that not all questions are always asked.


dodoels (approved):

arekmano (commented):
Binary is included in the repository

PR #361, opened 2018-02-26
[#350] additional indemnity money

Samuel-Campbell:
1) removed anti-regex because they are not used properly
1.1) Example: not_violent is not set to 1 if the violent regex does not match. However the NLP does so. Therefore the results are bad.
2) improved tenant pays landlord by changing loss function from mean_absolute_percentage_error to mean_squared error
3) Finished additional indemnity money
4) tested with and without using average values in a vector --> both yield very similar result during testing. Therefore using average column values is better when missing data
5) Each regressor now gets tested after training
6) METRICS:
**Tenant Pays Landlord:**
INFO: Size of dataset: 26594
INFO: R2: 0.70
INFO: Explained Variance: 0.70
INFO: Mean Absolute Error: 492.03
INFO: Mean Squared Error: 1362642.05
**Additional Indemnity Money:**
INFO: R2: 0.78
INFO: Explained Variance: 0.78
INFO: Mean Absolute Error: 471.99
INFO: Mean Squared Error: 831383.89

6) I think the error is caused by the date regex. It's probably not working well


arekmano (approved):
LGTM

PR #363, opened 2018-03-01
[#345] Ensures all Python-based services uniformly use Flask as opposed to Gunicorn

lancelafontaine:
[#345]

- [x] Ensures all Python-based services uniformly use Flask as opposed to Gunicorn

Reasons for this change:
- @Vynny previously had issues with Gunicorn's multiple workers.
- Logging output did not behave as expected.
- Minimize differences in Dockerfiles for Python-based services
- Speed is not a concern at this stage, so offers no benefit 
- One less dependency to download at build time

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -4,7 +4,7 @@ services:
   ml_service:
     ports:
       - 3001:3001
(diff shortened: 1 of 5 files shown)


arekmano (approved):
LGTM

Issue #263, opened 2018-01-12
Handwriting Analysis  for OCR


lancelafontaine:
*Description*
As a user, I would like to recognize words from handwriting for use in analysing leases.

*Scope of Work*
- [x] Investigate solutions for character recognition
- [x] Implement proof of concept for recognizing handwritten words in leases.

*Demo requirement*
1. The user submits an image of a written word
2. The system returns the recognised word

arekmano:
Related research paper:
http://ieeexplore.ieee.org/document/7424344/

arekmano:
Fairly certain this will involve some form of machine learning to classify handwritten digits/letters

Seems Tesseract does not do this out of the box. (https://stackoverflow.com/questions/39556443/using-tesseract-for-handwriting-recognition)

A possible implementation is using a CNN to do the handwritten character -> recognized char.

This would require us to implement some form of handwritten word to handwritten character splitting algorithm.

naregeff:
Some work was done; a lot more is required to deliver value.

PO changed priority, story relegated to backlog, to be picked up in later sprints.

lancelafontaine:
A proof of concept was created for handwriting analysis and can be seen in merged PR #372

lancelafontaine:
The remainder of the work would involved detecting word boundaries, and integrating the work into the task service API.

lancelafontaine:
The remainder of the OCR work required (asides from integrating the proof-of-concept in this issue) is described in issue #373.

PR #366, opened 2018-03-02
[#263] Fixes Keras Conv2D model error

lancelafontaine:
[#263]

```
ValueError: Negative dimension… size caused by subtracting 3 from 1 for 'conv2d_1/convolution' (op: 'Conv2D') with input shapes: [?,1,28,28], [3,3,28,32]
```

The `weights_epoch` variable was also undefined.

### DIFF
--- a/src/task_service/handwriting_recognition_ml.py
+++ b/src/task_service/handwriting_recognition_ml.py
@@ -61,17 +61,17 @@ def norm_input(x): return (x-mean_px)/std_px
 def create_model():
     model = Sequential([
         Lambda(norm_input, input_shape=(1,28,28), output_shape=(1,28,28)),
-        Conv2D(32, (3,3)),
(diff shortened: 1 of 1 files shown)


arekmano (approved):

src/task_service/handwriting_recognition_ml.py
```diff
@@ -100,6 +100,7 @@ def create_model():
 eval_batch_size = 512
 num_iterations = 1
 num_epochs = 1
```
Samuel-Campbell, line 102: not blocker, but those values are super small. they will need to be increased in the future

Samuel-Campbell (approved):

src/task_service/handwriting_recognition_ml.py
arekmano: yep. just a PoC. that img -> character works

Issue #37, opened 2017-10-04
Legal Definitions


naregeff:
*Description*
- As a layman, I want the chatbot to  provide legal definitions anytime it uses a legal term

The chatbot may from time to time use terms that are beyond the understanding of a regular citizen. It is important that these terms have definitions that can be gleaned by mousing over, or even by asking the bot what it means.

*Scope of Work*
- Add a dictionary of all possible legal terms to the system
- Add a tooltip hover for complex legal terms
- Add complex word outlining on the back-end

*Acceptance Criteria*
- Given a simple interaction, with the system asking the question: "Are you a tenant or a landlord?", both "tenant" and "landlord" words should clearly be outlined as hover-able. Upon mouse hover, their definitions should be shown to the user.

naregeff:
Verified, Priority medium.

arekmano:
**Attached are all the responses that the chatbot currently says:**


No. There is no allowable delay after signing a lease.

There are many recourses available to the tenant: rent reduction; rent deposit; authorization of the Régie to do himself the repairs; resiliation of the lease; Damages; order to force the landlord to make the repairs. A decision must be rendered by the Régie du logement allowing one of these recourses to be applied.

If you are suffering harassment, find someone to possibly serve as a witness testimony and document the events with the aid of an audio-visual device (ie: cell phone).

You should file an immediate demand letter to your landlord demanding for the harassment to cease and a possible lease resiliation or compensation for moral damages.

If your landlord is not receptive, you may file at the Régie du logement for a hearing. The Commission des droits de la personne et des droits de la jeunesse can also give you aid.

Be aware that if you leave your dwelling, you still need to pay unless the regie or the landlord allows you to resiliate your lease.

If your landlord fails to answer the phone or does not offer help in an emergency, call a professional immediately.

If your landlord refuses to refund you for the emergency service you paid for, write a demand letter informing the landlord that you need to be compensated and file for a hearing at the Regie du logement if you are not reimbursed.

If authorities, the insurance company or the Regie du logement find that the damage to the apartment was caused by your neglect, you may face financial repercussions. If not, any financial loss during the emergency will be compensated for including temporary accommodations.

If you are being abused by a spouse or a loved one, vacate the dwelling immediately and stay with friends, relatives, or contact a local shelter. File a police report immediately.

Acquire the form for requesting an attestation for lease resiliation due to spousal violence and fill it out.

Bring the form to a commissioner of oaths and get it signed.

[...]

PR #379, opened 2018-03-09
[#37] Added legal definitions to difficult words

arekmano:
screenshot from 2018-03-08 21-58-39


arekmano:
Data obtained from https://developer.oxforddictionaries.com API

dodoels:
Use `<el-tooltip>`. http://element.eleme.io/#/en-US/component/tooltip 
All the works can be done with one line of code :)

arekmano:
Doesn't work. Since we are inserting the HTML elements programmatically, vuejs does not compile the templates to HTML, so any vuejs component will not work. A vanilla javascript solution is needed for this story. 

https://vuejs.org/v2/guide/syntax.html#Raw-HTML
>[...] will be replaced with the value of the rawHtml property, interpreted as plain HTML - data bindings are ignored.

src/web_client/src/components/Chat.vue
```diff
@@ -176,8 +178,10 @@ export default {
         response => {
           // 1.1 save chat history in local and parse the sentences
           this.chatHistory.history = response.body.messages
+          var self = this
```
dodoels: just use `this`. It is not a good practice to use `self` in ES6.

src/web_client/src/components/Chat.vue
```diff
@@ -222,6 +226,21 @@ export default {
...
+        .split(' ')
+        .map((word) => {
+          if (Constants.difficult_word_definitions[word.toLowerCase()] !== undefined){
+            return '<span class="hoverable" title="' + Constants.difficult_word_definitions[word.toLowerCase()] +'" style="background-color: #f5af5380;border-radius: 3px;padding: 2px;">' + word + '</span>'
+          }
+          return word
+        })
+        .join(' ')
+    },
+    hoverWord (word) {
```
dodoels, line 239: remove dev code

src/web_client/src/components/Chat.vue
```diff
@@ -222,6 +226,21 @@ export default {
     }
   },
   methods: {
+    addHover (text) {
+      console.log(text)
+      return text
+        .split(' ')
+        .map((word) => {
+          if (Constants.difficult_word_definitions[word.toLowerCase()] !== undefined){
+            return '<span class="hoverable" title="' + Constants.difficult_word_definitions[word.toLowerCase()] +'" style="background-color: #f5af5380;border-radius: 3px;padding: 2px;">' + word + '</span>'
```
dodoels, line 233: can the styling be specified in scss file instead?

src/web_client/src/components/Chat.vue
```diff
@@ -250,6 +269,8 @@ export default {
       }
       // 2. set current zeus response
       let zeusResponseText = conversation.message || conversation.text
+      var self = this
```
dodoels: same here

dodoels (approved):
Seems like this is the easiest but not the elegant way to do it

src/web_client/src/components/Chat.vue
arekmano: won't "this" reference another item within an anonymous function?

src/web_client/src/components/Chat.vue
arekmano: no, because webpack will compile the scss, by adding additional specifications (i.e. data-XXXX fields) and these cannot be applied to v-html contents.

src/web_client/src/components/Chat.vue
arekmano: fixed

src/web_client/src/components/Chat.vue
arekmano: fixed

src/web_client/src/components/Chat.vue
arekmano: looked up docs; arrow function binds "this" automatically. fixed.

Ref:
http://exploringjs.com/es6/ch_arrow-functions.html#sec_traditional-functions-bad-non-methods

PR #384, opened 2018-03-10
382/outdated models

Samuel-Campbell:
**bug**
Not all models were updated on CJL

1) removed unused pip installs
2) added new command: -train --all  
2.1) This trains classifier, regressor and similarity finder
3) updated init.py with new models


dodoels (approved):
LGTM

Vynny (approved):

PR #385, opened 2018-03-10
43/model metrics

Samuel-Campbell:
1) Created a metrics binary
2) Next PR will be to create an endpoint to get these metrics
3) Purpose is to display some of them on the dashboard. I assembled as many numbers as possible

```
        {
            'data_set':{
                'size': 5000
            },
            'regressor':{
                'regressor name':{
                    'std': 4,
                    'mean': 5,
                    'variance': 42,
                    'mean_fact_vector': [3, 1, 5, 6, 2]
                }
            },
            'classifier':{
                'classifier name':{
                    'prediction_accuracy': 0.92,
                }
            }
        }
```


dodoels:
For the endpoint, we only need the non-technical metrics (like, our system is 96% accurate). Users don't know what f1 is, so returning such data will not have user value.

Samuel-Campbell:
@choitwao I thought we agreed that I should return as many statistics as possible and you guys choose which ones you want

dodoels:
@Samuel-Campbell more is better yes, but if we know for fact that these data won't be present anywhere on the client side in the future then transferring these data only decrease our system's performance (i.e. requesting 1 metrics, but returning 10, which x10 size data)

dodoels:
@Samuel-Campbell we do need the data for each numerical value to create the bell curve tho, so those data will be used to calculate/plot the curve. But for the F1 score stuff, we don't really need it. UNLESS we need to show them for DEMO purpose (like show it to CJL)

Samuel-Campbell:
@choitwao removed f1, precision, recall

Vynny (approved):
The structure of the return json might need to be changed, but this is good for now. We'll see how the story develops.

Issue #236, opened 2017-12-23
PO STORY: Establish platform on various social medias to acquire FAQ from their users


arekmano:
*Description*
As a developer, I would like obtain common Quebec tenant/landlord questions by searching online 

*Scope of Work*
- [x] Determine what online sources to consider
- [x] Establish criteria for determining type of user demand
- [ ] Collect and report structured data

*Demo requirement*
1. Define how and where we will record this data 
2. Ensure data is easily-accessible by whole team

*Acceptance Criteria*
- The team has clearly-labeled data on the types and frequencies of questions that users ask about tenant/landlord law

ghost:
Visited Hojo January 8th 2018 at 2:15pm
Summary: The HOJO Coordinator will aid us with any inquiry we might have about the law **in practice**. Quick answers and very accessible.
hojocoordinator@csu.qc.ca 

Note: Do not use the old HOJO FAQ, use [LikeHome](http://likehome.info/actions/). They do not posses a database with FAQ but registers the most common issues on this website.


"Spotted" Facebook pages January 8th 2018 at 2:30pm
Also talked to owner of "Spotted" pages on Facebook. He will let us advertise on his pages.

ghost:
Talked to the Managing Editor of The Concordian Jan 9th 2018 at 2:30pm
Summary: He's passing us to one of his staffs to see who'd like to pick it up.
managing@theconcordian.com 

Possible interview.

naregeff:
need to have a little meeting to figure it all out. work in progress.

lancelafontaine:
Moving to Iteration 8 and adding the Blocked label until we finalize all the marketing texts and finish up all tweaks to the sign up page (see bugs in Iteration 8).

arekmano:
This user story has been taken over by the stakeholder, as he will assume marketing work that is related to this project.

PR #387, opened 2018-03-11
[#236] Removes task_service from build until it is needed again

lancelafontaine:
[#236]

- [x] Removes task_service from build until it is used at which point it can be reintroduced
- [x] Removes './' in docker-compose configuration for specifying the path of a Dockerfile


arekmano (approved):
lgtm

Issue #43, opened 2017-10-09
Dashboard Reporting of Results


arekmano:
*Description*
- As a user, I would like claim predictions/results to be displayed in a dashboard report

*Scope of Work*
- [x] Create a high fidelity UI design of result dashboard, with result components
- [x] Create back-end API calls to support dashboard creation
- [x] Implement dashboard UI
- [x] prediction accuracy
- [x] standard deviation + variance of ints
- [x] ml end points

*Acceptance Criteria*
- User completes an interaction with the system. The prediction outcome dashboard is displayed to them.
-- Display questions with most weight for determining answer(s).
-- If answer has value other than boolean, represent answer as range (bell curve).
-- If damages, ventilate into categories whenever possible.
-- If fees are to be awarded, separate legal fees from damages.
-- If red flags in parsed lease, indicate in a separate section.
-- 5 similar decisions -> Include snipets on hover.
-- Field that shows average delay at RLQ for that type of dispute.
-- Icons/links for contacting lawyer, mediation services, etc.

naregeff:
Verified. Priority medium.

dodoels:
@Vynny for this particular issue, I suggest the api returns attributes for different outcome predictions, something like 
```
"response": "Here is my prediction",
"report": true,
"legal_fee": 500,
"lease_termination": true
```

Vynny:
I think there should be a separate domain object for a report. Such that there is a one to many relationship between Conversation and Report. This would allow us to show them a dashboard for every prediction, or at least every prediction for a claim category.

dodoels:
@Vynny how do you handle the transition between the end of the conversation and generating the report?

Vynny:
Well I figure the report can be generated at any time, I need a bit more details about whats expected out of this dashboard to know for sure what to return.

dodoels:
Will create hig-fid mock-up by friday

naregeff:
**Title of each issue identified**

- Display questions with most weight for determining answer(s).
- If answer has value other than boolean, represent answer as range (bell curve).
- If damages, ventilate into categories whenever possible.
- If fees are to be awarded, separate legal fees from damages.
- If red flags in parsed lease, indicate in a separate section.
- 5 similar decisions -> Include snipets on hover.
- Field that shows average delay at RLQ for that type of dispute.
- Icons/links for contacting lawyer, mediation services, etc.

arekmano:
[...]

PR #390, opened 2018-03-12
[#43] probabilities

Samuel-Campbell:
1) Added probabilities to predictions
2) Model now trains using probabilities = True --> increases training time
3) ReadMe updated with change in end point
4) Probabilities only on Classifiers obviously.

Only once approved
--------------------
- [ ] upload new models on cjl


dodoels (commented):
Why do we need a list of probability ?

Samuel-Campbell:
Everyone telling me mixed stuff but friday we discussed this. When we do a prediction we wanted to give the user it's probability. Answering more questions will increase probability or change the outcome

Samuel-Campbell:
Also what was said: "just give us back as many numbers as you can"

arekmano (approved):
looks good

PR #395, opened 2018-03-15
[#43] Renders normal curve data for any ranged fact value in dashboard using D3.js

lancelafontaine:
[#43]

- [x] Renders normal curve data for any ranged fact value in dashboard using D3.js

One issue that this PR overlooks is the fact that ranged fact data can possibly not fit a normal distribution. This is an issue we will need to address in the future.

### DIFF
--- a/src/web_client/src/components/Sidebar.vue
+++ b/src/web_client/src/components/Sidebar.vue
@@ -240,11 +240,11 @@ export default {
                     .attr("transform", "translate(" + MARGIN_SIZE + "," + MARGIN_SIZE + ")");
 
                 // Create standardized normal distribution data
-                let data = generateData(0, standardDeviation / mean);
-                let verticalValueData = generateVerticalData((value - mean)/standardDeviation);
-                let verticalMeanData = generateVerticalData(0);
-                let verticalLine = createLine();
-                let bell_line = createLine();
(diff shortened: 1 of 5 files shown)


lancelafontaine:
image

src/web_client/src/components/Sidebar.vue
```diff
@@ -216,6 +208,137 @@ export default {
             this.$localStorage.remove('usertype')
             this.$router.push('/')
         },
+        createBellCurves() {
```
dodoels, line 211: 1. use `this` instead of `self`
2. wrap this function in a vue component instead

lancelafontaine: 1. Using `self` is necessary to preserve original context, look at line 221
2. don't know how much sense it makes as a component, since everything is done by DOM manipulation :(. The component would have 1 div.

dodoels: Can you test the code?

lancelafontaine: Most new functions have been tested in 3ace85e96813467a635ce939910adb80cde83bf7

Samuel-Campbell (approved):

dodoels (approved):
LGTM, just make sure the slider works for multiple numerical outcomes and display corresponding d3 graph

Issue #399, opened 2018-03-16
Calculate indemnity fee


arekmano:
*Description*
As a user, I make a claim about non payment which predicts that I am eligible for additional indemnity money. The system returns to me how much indemnity I can receive which is a value separate from the total money received.

*Scope of Work*
- [x] Change additional_indemnity_fees regex to boolean
- [x] Tag precedents
- [x] Train model
- [x] Upload model to CJL
- [x] Add static formula to regressor

*Demo requirement*
1. The user does x
2. The system does y
n. Rest of a use case of a flow chart

*Acceptance Criteria*
- The user is able to x
- The user sees y
- The user hears z

dodoels:
Here is what we know:
- Based on http://legisquebec.gouv.qc.ca/en/ShowDoc/cs/A-6.002 section 28, the interest applies every quarter. 
- Assume the tenant hasn't pay for 22 months, which means he hasn't paid for 7 quarters (without remainder, since the payment is calculated on the first day of the third month of each quarter) -> `n = 7`
- The annual interest rate is at 6%, therefore the quarter interest rate = Portion of a quarter of a year * Annual interest rate = 91/365 * 0.06 = 0.01496 -> `i = 0.01496`
- Assume the rent is $1000. Because the interest rate applies every quarter, the basic value/portion for the interest rate to be applied = rent * 3 months per quarter = 1000 * 3 = 3000 -> `PMT= 3000`

Therefore, since every 3 months/a quarter the tenant owns another $3000, we use future value formula to calculate the total amount, and we can subtract the interest part out
- The total owing = PV(1+i)^n + (PMT/i)((1+i)^n-1) = 0+(3000/0.01496)*((1+0.01496)^7-1) = 21966.33
- The total interest = total owing - quarter rent * number of quarter = 21966.33 - (3000*7) = $966.33
- note that in our case, the present value in the formula is 0. Imagine you are the landlord 22 months ago, the tenant didn't owe you anything, so the balance was 0. Three months after the formula starts applying.

dodoels:
To sum up:
- n = total months of owing / 4
- i =  0.01496 (fixed)
- PMT = rent * 3
- PV = 0
- total owing = PV(1+i)^n + (PMT/i)((1+i)^n-1) 
- total interest = total owing - PMT*n

PR #413, opened 2018-03-22
[#399] removed fraudulent demand regex

Samuel-Campbell:
- [x] Remove demand regex (bonus task)
- [x] Change additional_indemnity_money to boolean
- [x] Removed additional indemnity money regressor
- [x] Test pipeline from beginning to end (make sure nothing breaks) - using sample of 1000
- [x] Removed outcomes with low occurrence 

1- Removing demand regex makes tagging quicker. They are not used anyways
2- Additional indemnity money will now become a boolean since we will be calculating it's value using a formula that we know of
3- Had to remove outcomes with low occurrence --> delcares housing inhabitable. These outcomes cannot be implemented without a custom multi class svm wrapper. Right now the default one from sklearn doesn't always work. Since the training and testing sets are randomly split AND the outcome occurrence is very low THEN there is a chance that there are 0 values for a given outcome in the training set. If this happens then no classifier is created. You will get null pointer errors from this in the future.


src/ml_service/data/test/regex/orders_landlord_notify_tenant_when_habitable.txt
```diff
@@ -1 +0,0 @@
-ORDONNE au locateur sous toute peine que de droit, d'aviser la locataire dès que le logement sera redevenu propre à l'habitation
```
TaimoorRana, line 1: This should pass. Why are we removing this?

src/ml_service/feature_extraction/post_processing/regex/misc/regex_lib_helper_test.py
```diff
@@ -11,4 +11,5 @@ def test_regex_finder(self):
         self.assertTrue('tenant_continuous_late_payment' in regex_lib_helper.regex_finder(sentence))
 
     def test_get_regexes(self):
-        self.assertIsNotNone(regex_lib_helper.get_regexes('demand_lease_modification'))
+        regex_lib_helper.get_regexes('additional_indemnity_money')
+        self.assertIsNone(None)
```
TaimoorRana: will never fail

src/ml_service/model_training/regression/multi_output/multi_output_regression.py
```diff
@@ -78,7 +74,7 @@ def predict(self, facts, outcomes):
             if outcomes[i] == 1:
                 column_name = MultiOutputRegression.classifier_labels[i][0]
                 if column_name == 'additional_indemnity_money':
-                    outcomes[i] = AdditionalIndemnityMoney().predict(facts)[0][0]
+                    outcomes[i] = 1 # placeholder
```
arekmano, line 77: ?

arekmano: will the business logic be on the ML Service side? If not, we can remove this attribute completely

arekmano (approved):
Looks good. Minor change

src/ml_service/data/test/regex/orders_landlord_notify_tenant_when_habitable.txt
Samuel-Campbell: i removed the regex

src/ml_service/model_training/regression/multi_output/multi_output_regression.py
Samuel-Campbell: we need thijs to keep the app functional

src/ml_service/feature_extraction/post_processing/regex/regex_lib.py
```diff
@@ -12,220 +12,6 @@ class RegexLib:
...
-        ], "MONEY_REGEX"),
-        ("tenant_eviction", [
-            re.compile(
-                r".+" + LANDLORD_REGEX +
-                r".*(expulsion|éviction).*" + TENANT_REGEX + r"",
-                re.IGNORECASE
-            )
-        ], "BOOLEAN")
-    ]
-
```
dodoels, line 228: I feel like it might be useful for CJLs research. Maybe comment it out?

dodoels (commented):
I was gonna say system test before merging... but ok

PR #414, opened 2018-03-26
[#399] changed regex from days to months, finished additional indemni…

Samuel-Campbell:
1) Interest rate from article 1619  -- > Annual interest rate is 18% or 6% per trimester. Verified with online calculator
2) The additional indemnity money uses the following facts: monthly payment, months unpaid
3) Output is still calculated by multioutputregression although it doesn't use a regresson. This decision was made to change as little code as possible.

- [x] Re-Tag precedents
- [x] test in docker
- [x] Train model
- [x] Upload model to CJL
- [x] Add static formula to regressor


arekmano (approved):
LGTM. did the calculation match up with examples from the dataset?

Issue #401, opened 2018-03-17
DEV STORY: Verify Regular Expression for Dates


arekmano:
*Description*
There seems to be something wrong with date detection in precedents.

arekmano:
dev story is completed and merged

PR #419, opened 2018-03-27
[#401] date regex fix

TaimoorRana:
Added/updated regexes for dates

### DIFF
--- a/src/ml_service/feature_extraction/post_processing/regex/regex_lib.py
+++ b/src/ml_service/feature_extraction/post_processing/regex/regex_lib.py
@@ -9,8 +9,8 @@ class RegexLib:
     DEMAND_REGEX = r"(demand|réclam)(ait|e|ent|aient)"
     DATE_REGEX = r"(janvier|février|fevrier|mars|avril|d'avril|mai|juin|juillet|d'août|août|aout|septembre|d'octobre" \
                  r"|octobre|novembre|décembre|decembre)"
-    DATE_RANGE_REGEX = r"(?i)(\d{1,2})?(?:er|èr|ere|em|eme|ème)?\s?(\w{3,9}) (\d{4}) " \
-                       r"(?:a|à|au|et|et se terminant le) (\d{1,2})?(?:er|èr|ere|em|eme|ème)?\s?(\w{3,9}) (\d{4})"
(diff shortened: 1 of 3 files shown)


src/ml_service/feature_extraction/post_processing/regex/regex_lib.py
```diff
@@ -7,7 +7,10 @@ class RegexLib:
     TENANT_REGEX = r"locataire(s)?"
     LANDLORD_REGEX = r"locat(eur|rice)(s)?"
     DEMAND_REGEX = r"(demand|réclam)(ait|e|ent|aient)"
-    DATE_REGEX = r"(janvier|février|mars|avril|d'avril|mai|juin|juillet|d'août|août|septembre|d'octobre|octobre|novembre|décembre)"
+    DATE_REGEX = r"(janvier|février|fevrier|mars|avril|d'avril|mai|juin|juillet|d'août|août|aout|septembre|d'octobre" \
+                 r"|octobre|novembre|décembre|decembre)"
+    DATE_RANGE_REGEX = r"(?i)(\d{1,2})?(?:er|èr|ere|em|eme|ème)?\s?(\w{3,9}) (\d{4})?\s?" \
```
Samuel-Campbell, line 12: careful. We don't want the time of the lease but the time for which a tenant hasn't paid.

Samuel-Campbell (changes requested):
Regex will catch too many irrelevant things

src/ml_service/feature_extraction/post_processing/regex/regex_lib.py
TaimoorRana: Oh, I didn't know we needed it specifically for that. I'll change it

Samuel-Campbell: its possible we don't even need the regex after thinking about it. Check ml channel

TaimoorRana: I saw the channel but this way almost done, so I finished it.

TaimoorRana: if we ever need it, it will be there.

Samuel-Campbell (approved):
can you tag precedents and upload model to CJL when you got time?

TaimoorRana:
sure

PR #429, opened 2018-03-31
[#399] Monthly Rent Fact + Bugfix

Vynny:
[#399]

- [X] Changed type of fact tenant_monthly_payment from BOOLEAN to MONEY
- [X] Added new training data and bot question for tenant_monthly_payment
- [X] Fixed a small bug in nlp service's ml_service wherein it wouldn't send over facts of type DURATION_MONTHS

**Note: Requires DB reset to function properly**


Samuel-Campbell (approved):
Tested and worked properly

Issue #405, opened 2018-03-17
DEV STORY: Update microservice READMEs


arekmano:
*Description*
As an open source developer, I would like the microservices' READMEs to be current.

*Scope of Work*
- [x] Document newly added functionality into the microservice READMEs.
- [x] Create a troubleshoot ReadMe
- [x] Update Web Services ReadMe
- [x] Update NLP ReadMe
- [x] Update ML ReadMe
- [x] Update Backend Readme

ghost:
NLP Doc and Troubleshoot Doc ReadMe updated and merged to master.

arekmano:
Final Version of Docs: https://cyberjusticelab.github.io/JusticeAI/docs/rendered/

PR #443, opened 2018-04-13
[#405] updated readme

Samuel-Campbell:
updated readme

### DIFF
--- a/src/ml_service/README.md
+++ b/src/ml_service/README.md
@@ -1,67 +1,344 @@
 # Machine Learning Service
 
-## Installation Instructions
+## 1. Overview 
+The machine learning service is responsible for predicting the outcomes
+of a user's case. 
+
+Outcomes can either be categorized as either being True/False or by a numerical value. 
+Whether a given outcome is boolean or integer is evaluated by a human and then given to the
+system beforehand. Therefore, this sub-system makes use of both classifiers and regressors 
+to make predictions. The inputs for both the classifier the and regressor are the facts obtained
+by the user's inputs. An array of outcomes is then returned. 
+
+
+### 1.1 Data Representation
+The input and output data are all represented numerically despite having the
+potential to be boolean values. Below illustrates how values are treated:
+
+0 --> False / Null  
+1 --> True    
+(n > 1) --> True AND Numerical
+
+
+Numerical Values consist of:  
+- Dates / Time (in months)
+- Money (in $)  
+
+
+### 1.2 Facts / Input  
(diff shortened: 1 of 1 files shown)


ghost (approved):

src/ml_service/README.md
```diff
@@ -1,67 +1,356 @@
...
+
+**\*_regressor.bin**    
+Models used to predict regressive results
+
+## 3. Installation Instructions
 
 1. Add Cyberjustice Lab username as environment variables: <code>export CJL_USER={USERNAME}</code> either to your .bashrc or run it as a command
 2. Add Cyberjustice Lab password as environment variables: <code>export CJL_PASS={PASSWORD}</code> either to your .bashrc or run it as a command
 3. Run <code>pip install -r requirements.txt</code>
 4. Run <code>pip install -r requirements_test.txt</code>
```
ghost: requirements_test.txt recursively calls requirements.txt

Also I feel installation Instructions should be one of the first things in the markdown.

src/ml_service/README.md
```diff
@@ -611,26 +900,31 @@ None
...
----
 
-## Run Tests and Lints
+
+## 6. Run Tests and Lints
 
 ```
 export COMPOSE_FILE=ci
-./cjl up -d && ./cjl run ml_service
+./cjl down && ./cjl up -d && ./cjl run ml_service
```
ghost: why do you down it?
If it's up it should have already passed the tests.

src/ml_service/README.md
ghost: Also careful here, we are using python 3, pip has a chance of not pointing to pip3 depending on the python install order (e.i. dev installed python 2.7 after 3.6). Would recommend using pip3 exclusively to avoid errors.

src/ml_service/README.md
ghost: Also why run YOUR service after you "up" every service?
This doesn't make much sense to me

ghost: imo avoid "-d" as well so you can see the logs while running the terminal

ghost (changes requested):

src/ml_service/README.md
Vynny: Using -d is better because it runs it in the background and you can simply use ./cjl logs -f to view logs for all the services, or ./cjl logs -f service_name to view logs selectively for a particular service.

ghost (approved):

dodoels (approved):

PR #444, opened 2018-04-13
[#405] FE doc

dodoels:
@lancelafontaine I remove web client in the docker build config in dev environment, since there is no way to dev on FE with docker. Although CI and CD will not be impact.


arekmano (approved):
looks good

PR #456, opened 2018-04-17
[#0] ML Service Dockerfile Fix

Vynny:
Old dockerfile was copying binaries into container, but container already downloads binaries in init.py. This was causing some weird bugs.


Samuel-Campbell (approved):
lgtm

